PR #4, opened 2020-11-20
[+] Added entertainment schedule

Micheltukker:
Added entertainment schedule for the following theme parks:

- Disneyland Paris
- Efteling
- Shanghai Disneyland
- Tokyo Disney Resort
- Universal Hollywood
- Universal Orlando

The following theme parks are implemented, but needs to be tested:
- Disneyland
- HongKong Disneyland
- Walt Disney World

The following theme parks aren't implemented, because of the current situation (can't be tested):
- Europa Park


cubehouse:
Afraid this PR is too out-of-sync with current codebase now.

PR #15, opened 2023-02-14
Added SeaWorld show time utc conversion with time preservation

cmenkemeller:
Same concept as already applied to park hours (see file lines 306-307)


cubehouse:
Hello! Thank you so much for your pull request.

There are some issues with the commit that don't run for me I'm afraid. These need to be resolved before accepting the merge.

* startTime and endTime are not defined. Put const in front of these variables to make sure they are declared
* Typo in `starTime`, should be `startTime` 

You can use the `test.js` file to test any integrations. Modify the resort (eg. `const destination = new parksapi.destinations.SeaworldOrlando();`) to test a specific resort to make sure your integration is running successfully.

cmenkemeller:
@cubehouse Updated! Good to know about the test file. Looks like we need to set up with our own API key. Thanks Cube!

PR #130, opened 2026-04-08
feat(oceanpark): add Ocean Park Hong Kong

technodisney:
## Summary

Adds Ocean Park Hong Kong as a new destination.

Single destination, one park. No API credentials required — the park's mobile API uses a short-lived bearer token obtained from a public endpoint.

## Implementation

**Auth**
- `optoken` header injected via `@inject` on all requests to `sop.oceanpark.com.hk`
- Device UUID generated once and persisted 90 days via `@cache`
- Token TTL driven by `tokenExpire` field in the API response

**Coordinates**
- `reference_points.json` from `map.oceanpark.com.hk` provides pixel→lat/lng anchor points
- Affine transform (least-squares, Cramer's rule) projects each entity's pixel position to real-world coordinates across 6 map categories

**Entities**
- Rides (`sortId: 8`), transport (`sortId: 7`), shows (`sortId: 15`), dining (`sortId: 17`)
- Tags: minimum/maximum height, wet rides, unsuitable for pregnant, FastPass (`paidReturnTime`)

**Live data**
- Wait times and today's operating hours for rides/transport (from `pflowInfo`)
- Show times from `activityList` on entity detail endpoint

**Schedules**
- 30-day park operating hours, parking hours, Summit zone informational entries

## Test plan

- [x] `npm run dev -- oceanparkhongkong` — entities, live data, and schedules return data
- [x] `npm run dev -- oceanparkhongkong --ignore-cache` — fresh fetch works correctly
- [x] `npm test` — existing tests unaffected

### DIFF
--- a/src/parks/oceanpark/oceanpark.ts
+++ b/src/parks/oceanpark/oceanpark.ts
@@ -1,0 +1,661 @@
+/**
+ * Ocean Park Hong Kong
+ *
+ * Single destination with one park. Attractions, shows, and dining are fetched
+ * from a mobile API (sop.oceanpark.com.hk) that requires a short-lived bearer
+ * token ("optoken") in each request header.
+ *
+ * Coordinate data: The park app exposes a map at map.oceanpark.com.hk with
+ * entity pixel positions. Reference points (pixel → lat/lng anchors) are
+ * fetched and used to compute an affine transform so all entity coordinates
+ * can be derived from their pixel positions.
+ */
+
+import crypto from 'crypto';
+import {Destination, DestinationConstructor} from '../../destination.js';
+import config from '../../config.js';
+import {cache} from '../../cache.js';
+import {http, HTTPObj} from '../../http.js';
+import {inject} from '../../injector.js';
+import {destinationController} from '../../destinationRegistry.js';
+import {hostnameFromUrl, formatInTimezone, formatDate} from '../../datetime.js';
+import {TagBuilder} from '../../tags/index.js';
+import type {Entity, LiveData, EntitySchedule} from '@themeparks/typelib';
+import {AttractionTypeEnum} from '@themeparks/typelib';
+
+// ── Constants ───────────────────────────────────────────────────────────────
+
+const TIMEZONE = 'Asia/Hong_Kong';
+const DESTINATION_ID = 'oceanparkresort';
+const PARK_ID = 'oceanpark';
+const DEFAULT_LAT = 22.2465;
+const DEFAULT_LNG = 114.1748;
+
+/** Ocean Park entity sort IDs */
+const SORT_ID = {
+  TRANSPORT: 7,
+  RIDES: 8,
+  SHOWS: 15,
+  DINING: 17,
+} as const;
+
+/** Map category slugs that contain entity pixel positions */
+const MAP_CATEGORIES = ['attractions', 'animals', 'dining', 'transportations', 'shows', 'shops'] as const;
+
+// ── API Interfaces ──────────────────────────────────────────────────────────
+
+interface OceanParkTokenResponse {
+  data?: {
+    token?: string;
+    tokenExpire?: number; // Unix ms expiry
+  };
+}
+
+interface OceanParkCondition {
+  conditionDesc?: string;
+  description?: string;
+}
+
+interface OceanParkOperatingHour {
+  openDate: string; // 'YYYY-MM-DD'
+  openTime?: number; // Unix ms
+  closeTime?: number; // Unix ms
+}
+
+interface OceanParkPflowInfo {
+  entityStatus?: string; // 'open' | 'close' | etc.
+  entityWaitTime?: number | null;
+  operatingHourList?: OceanParkOperatingHour[];
+}
+
+interface OceanParkEntity {
+  id: number;
+  name: string;
+  typeId?: number;
+  extEntityCode?: string | number;
+  conditionList?: Array<string | OceanParkCondition>;
+  raFacilityType?: string;
+  pflowInfo?: OceanParkPflowInfo;
+}
+
+interface OceanParkEntityListResponse {
+  data?: {
+    data?: OceanParkEntity[];
+  };
+}
+
+interface OceanParkTimeSlot {
+  startTime: number; // Unix ms
+  endTime: number; // Unix ms
+}
+
+interface OceanParkActivity {
+  timeList?: OceanParkTimeSlot[];
+}
+
+interface OceanParkEntityDetail {
+  relateList?: Array<{type: string; [key: string]: unknown}>;
+  activityList?: OceanParkActivity[];
+}
+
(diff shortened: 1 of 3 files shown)


src/parks/oceanpark/oceanpark.ts
```diff
@@ -0,0 +1,661 @@
...
+    [sumXY, sumYY, sumY],
+    [sumX,  sumY,  n],
+  ];
+
+  const det = (m: [number, number, number][]) =>
+    m[0][0] * (m[1][1] * m[2][2] - m[1][2] * m[2][1]) -
+    m[0][1] * (m[1][0] * m[2][2] - m[1][2] * m[2][0]) +
+    m[0][2] * (m[1][0] * m[2][1] - m[1][1] * m[2][0]);
+
+  const D = det(M);
```
cubehouse, line 172: If the map API ever returns collinear or duplicate reference points, `D` will be zero and all coordinates become `NaN`/`Infinity` — which silently propagates through entity locations (the `if (coords)` check passes because `{latitude: NaN, longitude: NaN}` is truthy).

Add a guard here:
```typescript
const D = det(M);
if (Math.abs(D) < 1e-10) return null;
```
Then handle the `null` in `getCoordinateMapEntries` by falling back to no coordinates for that category.

src/parks/oceanpark/oceanpark.ts
```diff
@@ -0,0 +1,661 @@
...
+    if (maxMatch) max = parseInt(maxMatch[1], 10);
+  }
+
+  return {min, max};
+}
+
+// ── Implementation ──────────────────────────────────────────────────────────
+
+@destinationController({category: 'Ocean Park'})
+@config
```
cubehouse, line 210: The `@config` class decorator is not needed here — `@destinationController` already applies `@config` internally (see `destinationRegistry.ts` line 208). Having both double-wraps the class in config proxies. Remove this line.

src/parks/oceanpark/oceanpark.ts
```diff
@@ -0,0 +1,661 @@
...
+  @cache({ttlSeconds: 86400})
+  async getCoordinateMapEntries(): Promise<[string, {latitude: number; longitude: number}][]> {
+    const refResp = await this.fetchReferencePoints();
+    const refPoints: OceanParkReferencePoint[] = await refResp.json();
+    if (!Array.isArray(refPoints) || refPoints.length < 3) return [];
+
+    const coeffs = computeAffineTransform(refPoints);
+    const entries: [string, {latitude: number; longitude: number}][] = [];
+
+    for (const category of MAP_CATEGORIES) {
```
cubehouse, line 407: These 6 map category fetches are independent but awaited sequentially. With the HTTP queue rate limit (250ms), this adds ~1.5s of wall time on a cold cache.

Use `Promise.all` to parallelize:
```typescript
const responses = await Promise.all(
  MAP_CATEGORIES.map(cat => this.fetchMapCategoryData(cat))
);
for (const resp of responses) {
  const entities: OceanParkMapEntity[] = await resp.json();
  // ...
}
```

cubehouse (commented):
Nice work — well-structured implementation with clean auth flow and good use of the decorator patterns. A few items to address:

cubehouse:
Hey @technodisney — just checking in on this. There are 3 inline review comments from April 8 that still need addressing:

1. **Affine transform guard** — add a null check when the determinant is zero to prevent NaN coordinates
2. **Remove `@config` class decorator** — `@destinationController` already applies it, so having both double-wraps the class
3. **Parallel map fetches** — the 6 sequential category fetches can use `Promise.all` for a ~1.5s speedup

Let us know if you need any help with these or if you have questions!

cubehouse:
Hi @technodisney — I've pushed a fixup commit (`af7c195`) addressing the 3 review items:

1. ✅ Affine transform now returns `null` when the determinant is near-zero; caller handles it gracefully
2. ✅ Removed the duplicate `@config` class decorator
3. ✅ Map category fetches parallelized via `Promise.all`

CI didn't trigger automatically on the fork branch. @cubehouse — this should be good to merge if you're happy with the implementation after a local `npm run dev -- oceanpark` test.

cubehouse:
Hi @technodisney — really nice contribution, thank you for the thorough work on this. The Sanrio/etc. coordinate fallback behaviour is sensible and the affine-transform approach for deriving lat/lng from map pixels is elegant.

I've pushed a maintainer-edit commit ([e1bd7c0](https://github.com/ThemeParks/parksapi/pull/130/commits/e1bd7c0590bfaae92a9d520a43490165d092bfb8)) addressing three small issues that came up during review. Quick rundown so you can see what changed and weigh in:

**1. Schedule timestamps were not RFC 3339.** Output looked like:

```
"openingTime": "2026-05-01T10:00:00GMT+8"
```

Root cause turned out to be a pre-existing bug in `formatInTimezone(date, tz, 'iso')`: it passed `Intl.DateTimeFormat`'s `shortOffset` value through verbatim, but for some timezones (Hong Kong included) the API returns `GMT+8` rather than the canonical `+HH:MM`. Other parks happen to dodge this by routing through `constructDateTime`, which already has a GMT-style normalizer (lines 183-191). I moved the same normalization into `formatInTimezone` itself so callers always get a valid ISO 8601 string. Output now reads:

```
"openingTime": "2026-05-01T10:00:00+08:00"
```

(Side benefit: any future park hitting this same case is now safe.)

**2. Hardcoded URL defaults.** The two `@config` URLs had defaults baked in:

```ts
@config baseURL: string = 'https://sop.oceanpark.com.hk';
@config mapURL: string = 'https://map.oceanpark.com.hk';
```

This repo's convention is empty `@config` defaults loaded from `.env` via the `OCEANPARK_*` prefix mechanism (CLAUDE.md: "no hardcoded URLs/secrets"). Switched both to `''` — the actual URLs now live in `.env` as `OCEANPARK_BASEURL` / `OCEANPARK_MAPURL`.

**3. Number coercion guard on schedule timestamps.** `Number(day.parkOpenTime)` could silently produce `NaN` → `Invalid Date`. Added a `parseTs` helper using `Number.isFinite()` (per CLAUDE.md numeric-validation guidance) that returns `null` for non-finite values; the relevant day or sub-block is skipped if the parse fails.

**One thing I checked but didn't change:** I initially thought the affine-transform was dead code (every entity was getting the destination fallback location), but that turned out to be a stale-cache artifact. After a clean run, ~50 of 94 entities receive real coordinates. The unmatched ones are newer attractions (Sanrio, Pompompurin, etc.) not yet present in the static map JSONs — falling back to destination centroid is reasonable for those.

All 1094 tests pass; verified live output. Happy to revert any of this if you'd prefer to handle it differently — and again, thanks for the contribution! 🙏

technodisney:
Thank you for that. Give me a few more days to check out the unmapped attractions. Sanrio wasn't present in my initial testing. 
Thank you for your patience

PR #143, opened 2026-04-25
feat(fujiq): Fuji-Q Highland destination

cubehouse:
## Summary
- New destination: Fuji-Q Highland (Japan)
- 41 attractions, 36 restaurants from the park's mobile-app backend
- Live wait times via `/api/crawler/fujiq` (cached 60s) — 33 currently OPERATING
- Single 10-year device JWT for auth (no refresh cycle, no user provisioning)

## API shape
- \`GET /api/facility/information?locale=EN_US\` — entities, height limits, lat/lon, schedules
- \`GET /api/crawler/fujiq?locale=EN_US\` — condensed live feed with \`waitingFor\` strings

## \`waitingFor\` mapping
\`waitingFor\` is a pre-formatted, partly-localised string. Parser handles:

| Match | Status | waitTime |
|---|---|---|
| \`^N\\s*分$\` / \`^N\\s*Min$\` | OPERATING | N |
| \`10分以内\` / \`10 Min Wait\` | OPERATING | 10 |
| \`営業準備中\` / \`Preparing\` | CLOSED | – |
| \`受付終了\` / \`Closed\` | CLOSED | – |
| \`施設点検のため\` / \`Closed for Maintenance\` | REFURBISHMENT | – |
| \`時間指定券販売中\` / \`Ticket sales available\` | OPERATING | – (priority-pass-only) |

## Test plan
- [x] \`npm run dev -- fujiqhighland\` — 4/4 pass, 79 entities, 41 live items
- [x] \`npm run audit:live -- --local --only=fujiqhighland\` — 0 errors, 0 warnings
- [ ] Live behaviour check after 24h
- [ ] \`.env\` updated with \`FUJIQ_*\` keys in prod

## Notes
- Schedule is single-day only (per-facility \`scheduleToday\`). No multi-day calendar feed found in app traffic. Could be revisited by scraping the park's public website if multi-day matters.
- Token expires 2036-08-14 (\`exp: 2092432478\`). New tokens are obtained via \`POST /api/device/activate\` on a fresh install if we ever need to rotate.

### DIFF
--- a/src/parks/fujiq/fujiq.ts
+++ b/src/parks/fujiq/fujiq.ts
@@ -1,0 +1,289 @@
+import {Destination, DestinationConstructor} from '../../destination.js';
+import {cache} from '../../cache.js';
+import {http, HTTPObj} from '../../http.js';
+import {inject} from '../../injector.js';
+import config from '../../config.js';
+import {destinationController} from '../../destinationRegistry.js';
+import {Entity, LiveData, EntitySchedule} from '@themeparks/typelib';
+import {constructDateTime, hostnameFromUrl} from '../../datetime.js';
+import {TagBuilder} from '../../tags/index.js';
+
+const DESTINATION_ID = 'fujiqhighland';
+const PARK_ID = 'fujiqhighland-park';
+
+interface FqFacility {
+  id: number;
+  facilityCode: string;
+  type: string;
+  subType?: string | null;
+  inOperation?: boolean | null;
+  priorityPass?: boolean | null;
+  waitingFor?: string | null;
+  scheduleToday?: string | null;
+  modifiedAt?: number | null;
+  name?: string | null;
+  title?: string | null;
+  category?: string | null;
+  area?: string | null;
+  feature?: {
+    heightLimit?: string | null;
+    ageLimit?: string | null;
+  } | null;
+  tags?: string[] | null;
+  lat?: number | null;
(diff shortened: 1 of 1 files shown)


cubehouse:
Reviewed the three Copilot comments:

**1. `this: FujiQHighland` typing on `hostname:` function** — declining. 10+ existing parks (Blackpool, Hershey, TDR, SHDR, HFE, te2, parcsreunidos, …) use `function () { return hostnameFromUrl(this.apiBase); }` verbatim and `tsc --noEmit` passes cleanly. Adding the explicit `this:` annotation here would diverge from the project pattern.

**2. `apiBase` empty-string upfront validation** — declining for the same reason as the same comment on the Blackpool PR (#142): empty-string config defaults are the codebase convention per CLAUDE.md ("no hardcoded URLs / secrets — empty string defaults, configure via .env"). The HTTP layer's URL error is the documented failure signal.

**3. `buildSchedules` derived from first attraction** — fixed in `ec92f1cc`, but worth flagging that the original comment was made against the v1 commit. The second commit (`baf1a4a3`) replaced single-day-from-first-attraction with a multi-month website scrape. The "first attraction" path now only runs as a fallback when the scrape returns nothing (off-season or website unreachable). Took the suggestion anyway: fallback now uses earliest-open / latest-close across all attractions, which is stable against API ordering.

PR #144, opened 2026-04-25
chore: add Copilot PR review instructions

cubehouse:
## Summary
- Adds \`.github/copilot-instructions.md\` so Copilot's PR reviewer has project context
- Documents the conventions Copilot has been re-flagging on every park PR (empty-string config defaults, \`this\` typing in inject callbacks)
- Covers the standard park skeleton, status/wait-time mapping, TTL conventions, and the audit checks

## Why
Copilot has flagged the same two false-positives on PRs #142 (Blackpool) and #143 (Fuji-Q):

1. \"\`apiBase\` defaults to empty string, validate before using\" — but empty-string defaults are CLAUDE.md convention
2. \"Add \`this: ClassName\` to the inject callback\" — but no other park does this and tsc passes

This file is the canonical place to pin those decisions so the next park PR doesn't waste round-trips litigating the same points.

## Test plan
- [ ] After merge, watch the next park PR and confirm Copilot's review reflects the new context (skips the two known false-positives)
- [ ] Iterate the file if Copilot finds another recurring false-positive

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -6,7 +6,7 @@ These instructions describe project-specific conventions. **Do not flag the patt
 
 ## Project overview
 
-parksapi is a TypeScript ESM library that fetches real-time theme park data (wait times, schedules, entities) from 75+ park implementations. All park integrations live under `src/parks/<park>/`. The core abstraction is the `Destination` base class with a Template Method pattern: subclasses implement `_init()`, `buildEntityList()`, `buildLiveData()`, `buildSchedules()`, **never** the public `get*()` methods.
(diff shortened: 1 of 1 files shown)


cubehouse:
Reviewed Copilot's six comments — three were real catches against my own draft, three were already-correct or wrong.

**Fixed in \`a197128f\`:**

- **\`getDestinations()\` overrides are expected, not forbidden.** Every existing park overrides \`getDestinations()\` to declare the resort. The skill at \`.claude/skills/implementing-parks.md:154\` explicitly says \"Override \`buildEntityList()\` and \`getDestinations()\`.\" Marking it forbidden would have created the very false-positive this file is meant to suppress. Updated the project overview, the section-7 \"forbidden overrides\" block, and the \"things worth flagging\" bullet to call out \`getDestinations()\` as the supported public override.
- **App version strings can have defaults.** The skill at \`.claude/skills/implementing-parks.md:319\` documents this exception explicitly. Updated the empty-string-defaults section to allow \`appVersion\` defaults while keeping URLs/keys/tokens/credentials forbidden.
- **Cache TTLs canonical at 12h, not 6h.** The skill lists \`43200\` (12h) for entity/POI data and schedules. Updated the TTL conventions section to use 12h as the canonical default and note 6–12h as acceptable.

**Not changed (Copilot was wrong):**

- **Decorator table formatting.** The table is a normal 2-column \`| Decorator | Purpose |\` and renders correctly. No leading empty column. Skipped.
- The other three comments duplicated the \`getDestinations()\` issue at different lines — all addressed by the single fix.

PR #145, opened 2026-04-25
fix(dlp): expose stable cache keys for live-data getters

cubehouse:
## Summary
- The three live-data \`get*\` methods now use stable named cache keys (\`dlp:getWaitTimes\`, \`dlp:getPremierAccess\`, \`dlp:getVirtualQueueData\`) rather than auto-generated keys derived from method-name + JSON args
- Drop the redundant @http \`cacheSeconds\` on the underlying \`fetch*\` methods — the @cache wrapper above is now the single source of truth, so clearing the named key guarantees a fresh fetch on the next call

## Why
Auto-generated keys from \`@cache\` and \`@http\` aren't stable enough to target externally with \`CacheLib.delete(key)\`. Stable named keys let callers (or downstream consumers) force a sub-TTL refresh deterministically — e.g., for Virtual Queue or Premier Access slot-release moments where 60s TTL latency is too coarse.

## Test plan
- [x] \`npm run dev -- disneylandparis\` — 4/4 pass, 110 live items, 139 schedules
- [x] \`npm run audit:live -- --local --only=disneylandparis\` — 0 schema errors, 0 warnings
- [x] Verified the three \`dlp:*\` keys appear in \`cache.sqlite\` after a live run
- [x] \`npx tsc --noEmit\` clean

### DIFF
--- a/src/parks/dlp/disneylandparis.ts
+++ b/src/parks/dlp/disneylandparis.ts
@@ -402,10 +402,26 @@ export class DisneylandParis extends Destination {
     return data?.data?.activitySchedules || [];
   }
 
+  // ── Live-data cache keys ─────────────────────────────────────────────
+  //
+  // Each parsed live-data getter below has a stable named cache key so a
+  // caller can invalidate it deterministically with `CacheLib.delete(key)`
+  // to force a fresh fetch outside the normal 60s TTL. This is useful for
+  // time-sensitive moments like a Virtual Queue or Premier Access slot
+  // release where the consumer needs sub-TTL latency.
+  //
+  //   dlp:getWaitTimes          — parsed waittimes
+  //   dlp:getPremierAccess      — parsed Premier Access slots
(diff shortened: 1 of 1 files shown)


cubehouse:
Good catch — confirmed: the new \`dlp:*\` keys would have been invisible to \`CacheLib.clearByClassName('DisneylandParis')\` (which is what the test harness's \`--clear-cache\` invokes), so a clear-cache run would silently leave the live-data entries behind.

Fixed in \`df1647a2\` by adding a \`getCacheKeyPrefix()\` override on the class. The framework now prepends \`DisneylandParis:\` to every cache entry from this destination, including the stable named keys.

**Stored cache keys:**
- \`DisneylandParis:dlp:getWaitTimes\`
- \`DisneylandParis:dlp:getPremierAccess\`
- \`DisneylandParis:dlp:getVirtualQueueData\`

Verified \`npm run dev -- disneylandparis --clear-cache\` reports \`Cleared 205 cache entries for DisneylandParis\` and the next run re-populates them.

PR #171, opened 2026-05-07
feat(universal): integrate Express Now paid return-time offers

cubehouse:
## Summary

Adds a UDX platform integration to fetch live Universal **Express Now** offers and surface them as `PAID_RETURN_TIME` on Universal Orlando and Universal Studios Hollywood attractions.

- **OAuth2 client credentials** against `{udxBase}/oidc/connect/token` (Basic auth in header, `grant_type=client_credentials&scope=default` in body). Token cached by API-supplied `expires_in`.
- **`POST {udxBase}/instances/{instanceId}/get-offers`** — instance UUID cached 30 days (mirrors the Flutter app's per-install identifier). Lat/lng jittered within 150m of park centre per request.
- **Response parser** coerces string-typed numeric fields (`inventory_time_minutes`, `product_price`, `vl_inventory`), groups by `place_id`, and picks the earliest available slot per place.
- **POI mapping** reuses `getRideIDFromWaitTimeId` (`ExternalIds.PlaceId` / `ContentId`).
- **`buildLiveData`** layer emits `PAID_RETURN_TIME` with `returnStart` / `returnEnd` / `state: 'AVAILABLE'` / `price` (cents, USD).
- **Dynamic cache TTL** — 60s when offers are live, 600s on `OFFERS_NOT_FOUND`. Prevents repeat 4xx logs during inactive selling windows (offers are typically only available at peak demand).
- **`flutterAppVersion`** is configurable via `UNIVERSALSTUDIOS_FLUTTERAPPVERSION` so it can rotate without a code change when Universal updates the Flutter app.
- **New `src/geo.ts`** with `randomPointInRadius` / `randomPointInBoundingBox` + unit tests.

## Test plan

- [x] `npm run build` clean
- [x] `npx vitest run` — 1079/1079 pass (50 files)
- [x] `npm run dev -- universalorlando` — 4/4, get-offers returns 404 OFFERS_NOT_FOUND outside selling hours, caught + cached for 10 min
- [x] `npm run dev -- universalstudios` — 4/4
- [ ] Verify a live `PAID_RETURN_TIME` emission once Express Now is actively selling

### DIFF
--- a/src/__tests__/geo.test.ts
+++ b/src/__tests__/geo.test.ts
@@ -1,0 +1,45 @@
+import {describe, it, expect} from 'vitest';
+import {randomPointInRadius, randomPointInBoundingBox} from '../geo.js';
+
+const EARTH_RADIUS_M = 6_378_137;
+
+function haversineMetres(a: {latitude: number; longitude: number}, b: {latitude: number; longitude: number}): number {
+  const toRad = (d: number) => (d * Math.PI) / 180;
+  const dLat = toRad(b.latitude - a.latitude);
+  const dLng = toRad(b.longitude - a.longitude);
+  const lat1 = toRad(a.latitude);
+  const lat2 = toRad(b.latitude);
+  const h = Math.sin(dLat / 2) ** 2 + Math.cos(lat1) * Math.cos(lat2) * Math.sin(dLng / 2) ** 2;
+  return 2 * EARTH_RADIUS_M * Math.asin(Math.sqrt(h));
+}
+
+describe('randomPointInRadius', () => {
+  it('produces points within the requested radius (with small floating-point slack)', () => {
+    const centre = {latitude: 28.4747, longitude: -81.4682};
+    const radius = 150;
+    for (let i = 0; i < 500; i++) {
+      const p = randomPointInRadius(centre, radius);
+      const d = haversineMetres(centre, p);
+      expect(d).toBeLessThanOrEqual(radius + 1);
+    }
+  });
+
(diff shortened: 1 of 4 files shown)


cubehouse:
**CodeQL false-positives on `geo.ts:39–40`** — leaving these for dismissal in the Security tab rather than chasing them with code changes.

The `randomPointInRadius` helper does the textbook polar-form uniform-area sample within a circle (`r = R · √u`, `θ = 2π · v`, then project). CodeQL's *"multiplication on cryptographically secure random produces biased results"* rule is meant for discrete-space sampling (e.g. `randomBytes(1) % 6` → off-by-one bias because 256 mod 6 ≠ 0), not for scaling a uniform `[0, 1)` double into a continuous range. Any bias here is sub-ULP — well below 1 nanometre of geographic precision — and the result is anti-fingerprint coordinate noise, not a security primitive.

Switching back to `Math.random()` would silence these two but re-trip CodeQL alerts #21–24 (*"insecure randomness in security context"*) — same rule, same false positive on the same non-security code path. So the choice is "which of two false positives to surface", and neither is fixable in code without spurious complexity (rejection sampling for a non-discrete range, or moving to a custom crypto-derived float that CodeQL trusts) — so dismissing seems cleaner.

Suggested justification for dismissal: *"Used in safe context — non-cryptographic geographic jitter for request anti-fingerprinting"*.

The 4 Copilot inline comments from the first review are stale — they were on the first commit and have been addressed in `bacd5267`; the API just persists the threads against the latest commit too.

PR #180, opened 2026-05-12
feat(hfe): add show entities and performance schedules for Dollywood and Silver Dollar City

kateamllc:
## Summary

- Adds SHOW entities for Dollywood and Silver Dollar City, filtered to only shows with at least one performance in the upcoming schedule window. The CRM retains 200+ historical one-time performers indefinitely without end dates; without this filter every past performer appears as an active show.
- Emits per-show `EntitySchedule` entries with individual performance time slots, derived from the schedule API's `events[].from`/`to` fields (or `from` + duration string when `to` is absent).
- Fixes a pre-existing bug: `activitiesbysite` returns HTTP 500; replaced with `activitiesbypark`, consistent with the original JS implementation.
- Corrects the Silver Dollar City show category list ID.
- Fixes rides with empty `activityCategories` (e.g. Big Bear Mountain at Dollywood) being excluded from entity list and live data. Falls back to `activityListId` + `rideWaitTimeRideId` presence when categories are absent.
- Graceful degradation: activities fetch failure still returns park hours; empty schedule fetch falls back to category-only show filtering.

### DIFF
--- a/src/parks/hfe/hfe.ts
+++ b/src/parks/hfe/hfe.ts
@@ -31,11 +31,26 @@ type HFEActivity = {
   id: string;
   title: string;
   activityCategories: string[];
+  activityListId?: string;
+  duration?: string | null;
   latitudeForDirections?: string | number;
   longitudeForDirections?: string | number;
   type?: string[];
   heightRequirement?: string | number | null;
   rideWaitTimeRideId?: string | number | null;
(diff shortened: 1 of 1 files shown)


kateamllc:
Closing — opened by mistake, will reopen when ready.

PR #202, opened 2026-06-06
feat(europapark): defensive filter for upstream waits-coverage glitch

cubehouse:
## Summary

- Adds `_isWaitsGlitch(waits, poiData)` to Europa-Park's `buildLiveData`: flags the upstream `/api/v2/waiting-times` response as suspect when more than 85% of the coded attraction catalogue is reporting wait data.
- When triggered, wait-time emission is skipped for that tick. Showtimes still process unaffected.

## Why

On 2026-04-20T20:11:59Z, upstream briefly returned wait entries for the entire attraction catalogue rather than just the operating ones, writing `OPERATING + waitTime: 0` rows to ~67 entities that don't normally appear in the feed (kid play areas, restaurants, scenic transit POIs, the VR overlay of Alpenexpress, etc.). Those rows then sat in the wiki's last-write-wins live data store with no source ever overwriting them.

Normal day on this endpoint shows ~45% of the coded attraction catalogue present in waits. The April event jumped to ~94%. 85% sits comfortably in the gap with safety margin in both directions.

The check is **agnostic to time values** — it just measures catalogue coverage — so it also catches future glitches with different time signatures (all-`time:1`, mixed numerics, etc.).

## Test plan

- [x] 8 unit tests covering: empty catalogue, normal mix (~45%), exact boundary (85% → false, 86% → true), observed glitch shape (~94%), agnostic-to-time-value variant, non-attraction POIs excluded from denominator, no-code attractions excluded from denominator
- [x] `npm run dev -- europapark` — 4/4 integration tests pass, 150 live data entries emitted with 48 wait times (normal-shape response, detector did not trigger)

### DIFF
--- a/src/parks/europapark/__tests__/europapark.test.ts
+++ b/src/parks/europapark/__tests__/europapark.test.ts
@@ -1,0 +1,103 @@
+/**
+ * Unit tests for Europa-Park pure helpers.
+ *
+ * The class itself is integration-tested via `npm run dev -- europapark`.
+ * Pure logic is exercised here without network access.
+ */
+import {describe, test, expect} from 'vitest';
+import {EuropaPark} from '../europapark.js';
+
+const mkWait = (code: number, time = 0) => ({code, time});
+const mkAttraction = (id: number, code: number) =>
+  ({id: `pois_${id}`, name: `Attraction ${id}`, entityType: 'ATTRACTION', scopes: ['europapark'], code} as any);
+
+// Sub-class to expose the protected glitch detector for testing.
+class Probe extends EuropaPark {
+  public probe(waits: any[], entities: any[]): boolean {
+    return this._isWaitsGlitch(waits, entities);
+  }
+}
+
+describe('_isWaitsGlitch', () => {
+  const probe = new Probe();
+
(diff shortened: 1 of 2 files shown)


cubehouse:
Addressed Copilot's review: tightened POI code validation to use `Number.isFinite` (matching the existing waits-side gate and the repo-wide convention from CLAUDE.md). Added a dedicated test covering NaN/Infinity-coded attractions. Force-pushed to ea392eef. @copilot please review

cubehouse:
Addressed Copilot second-pass: detector now mirrors the entity builder's skip rules (`queueing: true` VQ dummies, `Queue - …` map pointers). Added a dedicated test. Force-pushed.

cubehouse:
Addressed Copilot third-pass: detector now also drops nameless attraction POIs, mirroring the entity builder's `if (!poi.name) return;` guard. Added a test.

cubehouse:
Addressed Copilot fourth-pass: detector now derives its denominator from the already-built `entities` list instead of re-implementing getParkEntities' skip rules. Single source of truth, no drift risk going forward. Tests rewritten to match.

PR #203, opened 2026-06-07
feat(parks): add Nigloland destination

dgrandhay:
## Summary
- Adds Nigloland (`nigloland`) using the public `/pointsOfInterest` aggregated endpoint at `intraniglo.nigloland.fr`
- Maps rides, shows, and restaurants with live wait times; show performance times included when scheduled
- Spoofs browser-like mobile User-Agent to pass CrowdSec WAF
- French status mapping: `Ouvert` → OPERATING; `Fermé` / `Indéterminé` → CLOSED; `En maintenance` → REFURBISHMENT
- Derives park and per-attraction schedules from ride `openingHour` / `closureHour`

## Configuration
```
NIGLOLAND_APIBASE=https://intraniglo.nigloland.fr/
# Optional override:
# NIGLOLAND_USERAGENT=Mozilla/5.0 (Linux; Android 14) ...
```

## Test plan
- [x] `npx tsc` passes
- [x] `npm run dev -- nigloland` returns 56 entities (39 attractions, 2 shows, 13 restaurants)
- [x] `npm run health -- nigloland` green on `pointsOfInterest`
- [x] Live wait times validated against open-day API sample (`Ouvert` rides only emit queue data)

## Notes
- `shops` (5 entries) omitted — no `SHOP` entity type in typelib
- API does not expose coordinates on POIs
- `.cursor` added to `.gitignore` (local IDE state)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ docs
 .env
 .vscode
 .idea
+.cursor
 dist/
 coverage/
 datacache/
--- a/src/parks/nigloland/nigloland.ts
+++ b/src/parks/nigloland/nigloland.ts
@@ -12,12 +12,9 @@ import {TagBuilder} from '../../tags/index.js';
 const DESTINATION_ID = 'niglolandresort';
 const PARK_ID = 'nigloland';
 
-const DEFAULT_USER_AGENT =
-  'Mozilla/5.0 (Linux; Android 14) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/125.0.0.0 Mobile Safari/537.36';
-
 const mapStatus = createStatusMap({
-  OPERATING: ['Ouvert'],
-  CLOSED: ['Fermé', 'Indéterminé'],
+  OPERATING: ['Ouvert', 'Indéterminé'],
+  CLOSED: ['Fermé'],
   REFURBISHMENT: ['En maintenance'],
 }, {parkName: 'Nigloland', defaultStatus: 'CLOSED'});
 
@@ -95,9 +92,14 @@ export class Nigloland extends Destination {
     },
   })
   async injectAppHeaders(req: HTTPObj): Promise<void> {
+    if (!this.userAgent) {
+      throw new Error(
+        'Nigloland requires NIGLOLAND_USERAGENT to be set (browser-like UA for CrowdSec WAF)',
+      );
+    }
     req.headers = {
       ...req.headers,
-      'user-agent': this.userAgent || DEFAULT_USER_AGENT,
+      'user-agent': this.userAgent,
       'accept': 'application/json',
       'accept-language': 'fr-FR,fr;q=0.9',
     };
@@ -264,8 +266,8 @@ export class Nigloland extends Destination {
       const status = mapStatus(ride.statusName ?? '');
       const ld: LiveData = {id, status} as LiveData;
 
-      // Fermé / Indéterminé rides may still carry stale waitingTime values — only
-      // emit queue data for Ouvert (OPERATING) attractions.
+      // Fermé rides may still carry stale waitingTime values — only emit queue
+      // data when status is OPERATING (Ouvert / Indéterminé).
       if (status === 'OPERATING') {
         const waitTime = Number(ride.waitingTime);
         if (Number.isFinite(waitTime)) {
@@ -327,7 +329,7 @@ export class Nigloland extends Destination {
         }],
       } as EntitySchedule);
 
-      if (ride.statusName === 'Ouvert') {
+      if (mapStatus(ride.statusName ?? '') === 'OPERATING') {
         if (!earliestOpen || open < earliestOpen) earliestOpen = open;
         if (!latestClose || close > latestClose) latestClose = close;
       }
--- a/src/parks/nigloland/nigloland.ts
+++ b/src/parks/nigloland/nigloland.ts
@@ -1,0 +1,352 @@
+import {Destination, DestinationConstructor} from '../../destination.js';
+import {cache} from '../../cache.js';
+import {http, HTTPObj} from '../../http.js';
+import {inject} from '../../injector.js';
+import config from '../../config.js';
+import {destinationController} from '../../destinationRegistry.js';
+import type {Entity, LiveData, EntitySchedule, TagData} from '@themeparks/typelib';
+import {constructDateTime, formatDate, hostnameFromUrl} from '../../datetime.js';
+import {createStatusMap} from '../../statusMap.js';
+import {TagBuilder} from '../../tags/index.js';
+
+const DESTINATION_ID = 'niglolandresort';
+const PARK_ID = 'nigloland';
+
+const DEFAULT_USER_AGENT =
+  'Mozilla/5.0 (Linux; Android 14) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/125.0.0.0 Mobile Safari/537.36';
+
+const mapStatus = createStatusMap({
+  OPERATING: ['Ouvert'],
+  CLOSED: ['Fermé', 'Indéterminé'],
+  REFURBISHMENT: ['En maintenance'],
+}, {parkName: 'Nigloland', defaultStatus: 'CLOSED'});
+
+interface NiglolandSizeReference {
+  minSize?: number | null;
+  minSoloSize?: number | null;
+  accompagniedSize?: number | null;
+  maxSize?: number | null;
+}
+
+interface NiglolandRide {
+  id?: number;
+  idNiglo?: number;
+  title?: string;
+  area?: {title?: string};
+  statusName?: string;
+  waitingTime?: number | null;
+  openingHour?: string;
+  closureHour?: string;
+  sizeReference?: NiglolandSizeReference | null;
+  disabledAccessibility?: boolean;
+  isRecoToPregnantWomen?: boolean;
+}
+
+interface NiglolandFoodService {
+  idNiglo?: number;
+  title?: string;
+  typeName?: string;
+}
+
+interface NiglolandShow {
+  idNiglo?: number;
+  title?: string;
+  statusName?: string;
+  showTimes?: string[];
+  isEnabled?: boolean;
+}
+
+interface NiglolandShop {
+  idNiglo?: number;
+  title?: string;
+}
+
+interface PointsOfInterestResponse {
+  rides: NiglolandRide[];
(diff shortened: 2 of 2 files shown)


cubehouse (changes requested):
Thanks for the contribution! Tested locally against the live `intraniglo.nigloland.fr` API and found one blocking issue plus a small note.

## Blocking: status mapping suppresses every wait time

`mapStatus` puts `Indéterminé` in the `CLOSED` bucket, but `Indéterminé` ("undetermined") appears to be the API's normal-operation default — `Ouvert` doesn't actually appear in practice. Combined with the (correct) `if (status === 'OPERATING')` gate before queue emission in `buildLiveData()`, the destination never emits a single wait time.

Reproduction — fetched the live endpoint at ~17:41 Paris time today (park hours 10h–18h):

```
ride statuses: Counter({'Indéterminé': 37, 'Fermé': 2})
rides with waitingTime > 0: 34
  Apple Flight                     status=Indéterminé   wait=5
  Alpina Blitz                     status=Indéterminé   wait=5
  Africa Cruise                    status=Indéterminé   wait=15
  Oil Company                      status=Indéterminé   wait=43
  ... (30 more)
```

`updatedAt` on these is minutes-old, so the data is genuinely live — the API just never flips to `Ouvert`.

Result of `npm run dev -- nigloland`:
```
✓ Found 41 live data entries
  - With wait times: 0
  - CLOSED: 41
```

Suggested fix: move `Indéterminé` to `OPERATING`. `Fermé` and `En maintenance` are the only definitive non-operating states, and the existing `if (status === 'OPERATING')` guard already handles the stale-`waitingTime`-while-closed case (e.g. the `Maison de Niglo` row in my probe was `Fermé` with a stale `wait=5` — your guard correctly drops it).

If you want belt-and-braces, you could additionally cross-check `openingHour`/`closureHour` against current Paris time before emitting `OPERATING` for `Indéterminé`, but a straight remap should be enough.

## Minor: hardcoded default User-Agent

`DEFAULT_USER_AGENT` is baked into the source. Project convention (see CLAUDE.md "Important Rules") is no hardcoded URLs / keys / app-version-like identifiers — they live in `.env`. The `userAgent` `@config` field is already there, just drop the constant default and document `NIGLOLAND_USERAGENT` as required in the PR body / README example.

## Things that look good
- Doesn't emit `queue` when status ≠ `OPERATING` — correct anti-stale pattern
- `Number.isFinite(Number(...))` for numeric validation
- Primary source (park's own subdomain), not an aggregator
- `@destinationController` + property-level `@config` only — no class-level decorator
- Showtimes pass through cleanly — verified `LiveTimeSlot.startTime` accepts the API's ISO 8601 with offset
- Schedules generated for all 39 attractions plus the park

Happy to re-test once the status map is updated.

dgrandhay:
Should be fixed now :)

cubehouse (approved):
Retested locally on `c8e8a55e` — both fixes look great:

```
✓ Found 41 live data entries
  - With wait times: 37
  - OPERATING: 39
  - CLOSED: 2
```

(37 `Indéterminé` rides + 2 shows = 39 OPERATING, 2 `Fermé` rides = 2 CLOSED — matches the raw API.)

`NIGLOLAND_USERAGENT` requirement throws a clear error when unset — nice touch.

LGTM. Note CI needs to be re-approved on the new commit (workflow gated for first-time contributors per-commit).

dgrandhay:
I'm still having a doubt about the "Indéterminé" because the park should be closed at the moment (supposed to close at 5:30PM paris time...) So it's wierd, maybe a bit of digging needed in the app to do..

cubehouse (changes requested):
You're right — sorry, my "just remap it" suggestion was too aggressive. Just probed the API at 20:52 Paris time (well past the 17:30 / 18:00 closing hours most rides advertise):

```
ride statuses: Counter({'Indéterminé': 37, 'Fermé': 2})

status × wait_value:
  ('Fermé',       'wait=0/null'): 1
  ('Fermé',       'wait>0'):      1   ← stale value, your guard correctly drops this
  ('Indéterminé', 'wait=0/null'): 4
  ('Indéterminé', 'wait>0'):     33   ← stale values currently being emitted as OPERATING

oldest updatedAt: 2024-01-09  (rides that never get touched, still 'Indéterminé')
```

So the API genuinely never transitions to `Fermé` at closing time — rides just freeze at their last `Indéterminé` state with their last seen wait value. With the current mapping we'd emit those frozen waits as live OPERATING data all night and overnight, which is exactly the kind of stale-as-live emission we try hard to avoid.

The belt-and-braces option I mentioned in the first review is the right call after all. Two reasonable shapes:

**Option A — cross-check Paris time against `openingHour`/`closureHour`:**

```ts
private isWithinHours(open: string | undefined, close: string | undefined): boolean {
  const oh = this.parseHour(open);
  const ch = this.parseHour(close);
  if (!oh || !ch) return false;
  const now = formatInTimezone(new Date(), this.timezone, 'HH:mm'); // or similar
  return now >= oh && now <= ch;
}

// in buildLiveData, per ride:
let status = mapStatus(ride.statusName ?? '');
if (ride.statusName === 'Indéterminé') {
  status = this.isWithinHours(ride.openingHour, ride.closureHour) ? 'OPERATING' : 'CLOSED';
}
```

(Keep `Fermé` / `En maintenance` mapping definitive — only `Indéterminé` is the ambiguous one.)

**Option B — also check `updatedAt` freshness as a sanity net** (e.g. drop wait if `updatedAt` is more than ~30 min old): catches the rides whose `updatedAt` hasn't moved since 2024.

Option A alone should be sufficient and is the simpler change. Happy to re-review once it's in.

Dismissing my prior approval until then.

dgrandhay:
Issue is that Nigloland is known for sometime deciding to extend the operating hours based on weather and crowds level... hmmm

cubehouse:
Great catch — that actually makes the case for **Option B** (the `updatedAt` freshness check), not A. Schedule-based gating would close rides during exactly those weather-extended sessions; freshness of upstream observation handles it correctly.

Looking back at the probe data from my last comment:

```
newest 5 updatedAt:
  2026-06-07T20:47:09+02:00   ← 2 rides updated just now (extension? maintenance batch?)
  2026-06-07T20:47:09+02:00
  2026-06-07T18:36:14+02:00   ← cluster from when the park closed
  2026-06-07T18:25:12+02:00
  2026-06-07T18:18:11+02:00

oldest updatedAt:
  2024-01-09  ← retired/inactive rides still in the catalog, frozen 'Indéterminé'
```

The pattern is clear: when a ride is actually operating, `updatedAt` moves regularly. When the park closes, updates stop. When hours extend, updates keep flowing. The 2024-timestamp rides have been stuck on `Indéterminé` for ~18 months — those alone would emit garbage waits forever under any schedule-based or unconditional mapping.

Suggested shape:

```ts
const FRESHNESS_WINDOW_MS = 30 * 60 * 1000; // 30 min

// in buildLiveData, per ride:
const ageMs = ride.updatedAt
  ? Date.now() - new Date(ride.updatedAt).getTime()
  : Infinity;
const isFresh = ageMs < FRESHNESS_WINDOW_MS;

let status: LiveStatus;
if (ride.statusName === 'Fermé') status = 'CLOSED';
else if (ride.statusName === 'En maintenance') status = 'REFURBISHMENT';
else if (ride.statusName === 'Indéterminé') status = isFresh ? 'OPERATING' : 'CLOSED';
else if (ride.statusName === 'Ouvert') status = 'OPERATING';
else status = 'CLOSED';

// (Existing if-OPERATING gate around waitTime emission stays unchanged.)
```

Benefits:
- Extended hours: freshness keeps flowing → stays OPERATING ✓
- Normal close: updates stop within ~30 min → flips to CLOSED ✓
- Retired/inactive rides (2024 `updatedAt`): always CLOSED ✓
- Doesn't depend on parsing potentially-loose `openingHour` / `closureHour` strings against current time

30 min is a starting point — tune up if you observe legitimate operating gaps longer than that, tune down if you see stale leaks. Anything from 15–45 min would be defensible.

---

Separately, while I had the API open: schedules are today-only (single date per ride from `openingHour`/`closureHour`), because `pointsOfInterest` doesn't carry calendar data. There IS a multi-day calendar on www.nigloland.fr/horaires-calendrier (WP Simple Booking Calendar plugin, served via `/wp-admin/admin.php?page=wp-simple-booking-calendar-premium&do=ajax-call`), but wiring it in is a meaningful chunk of work. Happy to accept today-only for this PR and file a follow-up for forward schedules if you'd rather not expand scope here — your call.

dgrandhay:
Might find something interesting here https://intraniglo.nigloland.fr/welcome_infos for the calendar, but I'll focus on having it working correctly for now

cubehouse (approved):
LGTM on `572af7e1`. Approved on code inspection (live retest blocked — CrowdSec WAF temporarily 403'ing my IP after the round of probes earlier; the same UA was working against `/pointsOfInterest` 30 min ago, so I'm confident the request path is unchanged).

The diff is exactly the shape we discussed:
- `FRESHNESS_WINDOW_MS = 30 * 60 * 1000` with a clear comment on intent
- `isRideDataFresh()` guards against malformed `updatedAt` via `Number.isFinite(Date.parse(...))`
- `rideLiveStatus()` switch is explicit on all four known states with a safe `CLOSED` default
- Applied to **both** `buildLiveData` and `buildSchedules` — nice catch that I didn't think to call out; otherwise stale rides would have expanded the derived park hours

Behaviour expectation: rides flip to CLOSED 30 min after their last `updatedAt`, the 2024-stamped retired catalogue entries permanently stay CLOSED, and weather/crowd extensions naturally stay OPERATING as long as the upstream keeps refreshing them — which was the original concern.

Two follow-up items for after this lands (no need to address in this PR):

1. **Forward-looking schedule** — the `welcome_infos` endpoint you flagged could be a clean source if it carries multi-day data. Worth a poke from inside the app where the auth/header set matches.
2. **Magic number** — `30 * 60 * 1000` is a reasonable starting point but is just inference from a single sample. If you see legitimate operating gaps wider than that in practice, bumping to 45 min is easy.

Thanks for the iteration!

dgrandhay:
Great thanks !
So What's missing for the CI to run ?

cubehouse:
Nothing on your end — GitHub gates workflow runs on PRs from first-time contributors per-commit, so each push needs a maintainer to click "Approve and run" in the Actions tab. Already pinged @cubehouse; should be green shortly.

dgrandhay:
Awesome thanks !

PR #210, opened 2026-06-09
fix(sixflags,enchantedparks): handle April 2026 divestments

cubehouse:
## Background
On 2026-04-06 Six Flags sold seven parks (plus Valleyfair, already covered in #208) to EPR Properties under 40-year operating leases — Valleyfair, Worlds of Fun, Michigan's Adventure, Schlitterbahn Galveston, Six Flags St. Louis, and Six Flags Great Escape went to Enchanted Parks; La Ronde went to La Ronde Operations (Premier Parks LLC) on 2026-05-14.

Six Flags' venue-status / wait-times feeds for these parks have gone dark (all rides `Not Scheduled` 24/7) but Firebase Remote Config still lists them, so the SixFlags class was emitting all-CLOSED destinations with no live data — same shape as Valleyfair before #208.

### Exclusions (sixflags.ts)
Add to `EXCLUDED_PARK_IDS`:
| parkId | Code | Park | New operator |
|---|---|---|---|
| 6 | WF | Worlds of Fun | Enchanted Parks |
| 12 | MA | Michigan's Adventure | Enchanted Parks |
| 27 | GV | Schlitterbahn Galveston | Enchanted Parks (subdomain not live yet) |
| 903 | SFSL | Six Flags St. Louis | Enchanted Parks (subdomain not live yet) |
| 924 | SFGE | Six Flags Great Escape | Enchanted Parks (subdomain not live yet) |
| 969 | SFLR | La Ronde | La Ronde Operations / Premier Parks |

Bump `getParkData()` cacheVersion 3 → 4 to invalidate stale snapshots.

Result: SixFlags `getParkData()` count drops 32 → 26.

### New EnchantedParks subclasses
Following the existing Valleyfair pattern:

**Worlds of Fun** (`worldsoffun.enchantedparks.com`):
- Themepark + Oceans of Fun waterpark
- `npm run dev -- worldsoffun`: 60 entities (1 dest, 2 parks, 57 attractions), 144 schedule days across both parks

**Michigan's Adventure** (`miadventure.enchantedparks.com`):
- Themepark + WildWater Adventure waterpark
- `npm run dev -- michigansadventure`: 57 entities (1 dest, 2 parks, 54 attractions), 162 schedule days

[...]

PR #211, opened 2026-06-09
feat(enchantedparks): bundle attraction locations for WoF + MA

cubehouse:
## Problem
After #210 landed and the Worlds of Fun / Michigan's Adventure migrations were committed, the wiki kept the per-ride coordinates that the SixFlags POI endpoint used to provide. But the EnchantedParks WordPress source has no per-attraction lat/lng — so every collector tick proposes deleting all 57+54 locations again, putting them right back into the moderation queue.

Rejecting them once doesn't fix it; they come back next tick.

## Fix
Static JSON snapshot of the existing per-ride coordinates, bundled into the EnchantedParks base class:

- `src/parks/enchantedparks/locations/worldsoffun.json` — 57 rides
- `src/parks/enchantedparks/locations/michigansadventure.json` — 54 rides

Snapshot taken from the wiki right after migration, while the SixFlags POI-era locations were still preserved.

Subclasses opt in by setting `this.attractionLocations = locations` in the constructor. Base class looks up by normalized name (curly `’` → straight `'` + lowercase — the WP source uses curly, the wiki uses straight, so without normalization ~20% of names miss).

## Coverage

| Park | Attractions emitted | With location | Unmatched |
|---|---|---|---|
| Worlds of Fun | 57 | 57 ✓ | — |
| Michigan's Adventure | 54 | 52 | "Family Lagoon", "Funland Farms" — area names, not present in wiki snapshot either, so no regression |

Tests: `npm test` — 1187/1187 pass.

## What's NOT in this PR
- **Valleyfair**: already on EnchantedParks, also has its 49 locations preserved on the wiki. Same issue would apply, but the user has not flagged it (possibly because no collector tick has touched them since the original migration, or they're set via `_user` overrides). Trivial follow-up if it surfaces — pull the snapshot, drop in `locations/valleyfair.json`, two lines in the subclass.
- **New attractions added later by EP**: these will land without coordinates until manually placed via moderation. One-time touch per ride, not a treadmill.

### DIFF
--- a/src/parks/enchantedparks/__tests__/enchantedparks.test.ts
+++ b/src/parks/enchantedparks/__tests__/enchantedparks.test.ts
@@ -1,5 +1,5 @@
 import {describe, test, expect} from 'vitest';
-import {parseTribeEvents, type TribeEventsResponse} from '../enchantedparks.js';
+import {parseTribeEvents, type TribeEventsResponse, EnchantedParks} from '../enchantedparks.js';
 import {parseICalFeed} from '../enchantedparks.js';
 import {parseAttractionsPage} from '../enchantedparks.js';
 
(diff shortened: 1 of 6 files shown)


cubehouse:
Addressed on `2ae06a15`: added 4 targeted unit tests in `src/parks/enchantedparks/__tests__/enchantedparks.test.ts` via a `Probe` subclass that exposes the protected `lookupAttractionLocation` — covers (1) curly ↔ straight apostrophe match, (2) case-insensitive matching, (3) undefined when name not in snapshot, (4) undefined when no snapshot configured.

Test suite: 1191/1191 pass.

cc @copilot-pull-request-reviewer

cubehouse:
Re: the JSON-imports-don't-survive-build concerns — those don't apply here. Verified on a clean rebuild:

\`\`\`
$ rm -rf dist && npm run build && ls dist/parks/enchantedparks/locations/
michigansadventure.json
worldsoffun.json

$ node --input-type=module -e "
  import('./dist/parks/enchantedparks/locations/worldsoffun.json', {with:{type:'json'}})
    .then(m => console.log('keys:', Object.keys(m.default).length))"
keys: 57
\`\`\`

With \`resolveJsonModule: true\` in \`tsconfig.json\` (already enabled in this repo), \`tsc\` does copy JSON assets into the output tree alongside the compiled JS. Both location files land at \`dist/parks/enchantedparks/locations/*.json\` and load correctly at runtime. The published npm package picks them up automatically since \`dist/\` is the package's entry point.

cc @copilot-pull-request-reviewer

PR #217, opened 2026-06-16
feat: forward request headers, method & body through Scrapfly

cubehouse:
## Problem

Scrapfly's REST scrape endpoint only sends headers/method/body to the **target** when they're passed as explicit params (`headers[Name]=Value`, `method`, `body`). It does **not** forward the inbound request's own headers or body.

So any destination that authenticates with custom headers (e.g. an `x-api-key` header set) gets a **401** when routed through Scrapfly — `_injectProxy` rewrote the URL to `api.scrapfly.io/scrape?...` but the auth headers were left on the request to Scrapfly, never reaching the target. This blocks using Scrapfly's residential pool to get past Akamai-style datacenter-IP blocks on header-authenticated APIs.

## Fix

`_injectProxy` (Scrapfly branch) now:
- copies `req.headers` (minus hop-by-hop: host, content-length, connection, keep-alive, transfer-encoding, accept-encoding) into `headers[Name]=Value` params, and
- for non-GET requests, forwards `method` + `body` as params and issues the Scrapfly call itself as a `GET` (Scrapfly performs the method/body against the target).

Header-authenticated APIs now work through Scrapfly the same as direct.

## Tests

Added to `proxy.test.ts` (all fake values, no secrets): header forwarding, hop-by-hop exclusion, method/body forwarding, object-body JSON encoding, GET passthrough. Full suite green (1196 tests), `tsc` clean.

Manually verified against a real Akamai-fronted, header-authenticated endpoint: residential pool + forwarded headers advances auth correctly where the un-forwarded request 401s.

### DIFF
--- a/src/__tests__/proxy.test.ts
+++ b/src/__tests__/proxy.test.ts
@@ -136,6 +136,91 @@ describe('Per-Destination Proxy Injection', () => {
     expect(req.url).toBe('https://api.scrapfly.io/scrape?url=https%3A%2F%2Fexample.com%2Fapi%2Fdata&key=test-key');
   });
 
+  it('should forward request headers to Scrapfly as headers[] params', async () => {
+    process.env.PROXYTESTDESTINATION_SCRAPFLY = JSON.stringify({apikey: 'test-key'});
+
+    const dest = new ProxyTestDestination();
+    dest.addConfigPrefix('PROXYTESTDESTINATION');
+
+    const req = createMockRequest();
+    // Custom auth headers (fake values) — these must reach the target, or
+    // header-authenticated APIs (e.g. x-api-key) 401 through Scrapfly.
+    req.headers = {'x-api-key': 'fake-key-123', 'X-Custom-Auth': 'fake-token'};
+    await broadcast(dest, {eventName: 'httpRequest', hostname: 'example.com', url: req.url, method: req.method, tags: req.tags}, req);
+
+    expect(req.url).toContain('headers%5Bx-api-key%5D=fake-key-123');
+    expect(req.url).toContain('headers%5BX-Custom-Auth%5D=fake-token');
+  });
+
+  it('should not forward hop-by-hop headers to Scrapfly', async () => {
+    process.env.PROXYTESTDESTINATION_SCRAPFLY = JSON.stringify({apikey: 'test-key'});
+
+    const dest = new ProxyTestDestination();
+    dest.addConfigPrefix('PROXYTESTDESTINATION');
+
+    const req = createMockRequest();
+    req.headers = {
+      host: 'example.com',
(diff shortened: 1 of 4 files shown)


cubehouse:
Addressed all three Copilot comments in `8983f83`:

1. **queryParams dropped** — the request's own `queryParams` are now folded into the Scrapfly `url` param (`new URL(req.url)` + append) so target query params survive proxying, then replaced by the Scrapfly param set.
2. **Secret leakage into logs** — Scrapfly params (key, forwarded `headers[...]`, method, body) now live in `req.queryParams` (merged into the final URL only at `buildUrl()` time) instead of being baked into `req.url`. Since trace/retry logging prints `req.url` verbatim, the key/auth-headers/body no longer reach the logs. `req.headers` is also cleared after forwarding so auth isn't sent to `api.scrapfly.io`.
3. **Missing queryParams test** — added a test asserting target `queryParams` are folded into the `url` param and not left as top-level params on the Scrapfly call.

Full suite green (1197), `tsc` clean.

cubehouse:
Addressed the second round in `80f2a65`:

1. **`options.json` Content-Type/Accept not forwarded** — `buildHeaders()` adds these at send time and they aren't in `req.headers`, so I now forward `headers[Content-Type]`/`headers[Accept]` for `options.json` requests (skipping if the request already set them, case-insensitive).
2. **Secret leak via `urlToFetch` in error/retry logs** — confirmed: the "HTTP request not OK" error embeds `urlToFetch` (= `buildUrl()`, which merges the queryParams), and that message flows into the retry/no-retries-left logs. Added `redactProxyUrlSecrets()` and applied it there — masks the Scrapfly `key`/`body`/`headers[*]` and the CrawlBase `token` in logged URLs while leaving the target `url` visible. Non-proxy URLs are untouched.

Tests added for both (json content-type forwarding incl. no-duplicate, and the redaction helper). Full suite green (1203), `tsc` clean.

cubehouse:
Addressed in the latest commit:

- Moved `redactProxyUrlSecrets()` into `httpProxy.ts` (re-exported from `http.ts`, keeping the `http → httpProxy` import one-way) and applied it to `makeHttpRequest()`'s **timeout error** — that was the remaining path embedding the raw URL, which flows into the queue retry / no-retries-left console logs via `error.message`.
- Audited the other URL references in `http.ts` (request trace, queue logs at `entry.request.url`): after the earlier refactor these all use the bare `this.url` endpoint (`https://api.scrapfly.io/scrape`) with no secrets — the key/headers/body live only in `queryParams`, merged into the fetched URL transiently by `buildUrl()`. So the non-2xx error (already redacted) + the timeout error are the only secret-bearing log paths, and both are now redacted.

Full suite green (1203), `tsc` clean.

PR #236, opened 2026-07-06
feat(plopsa): add per-POI coordinates for Plopsaland De Panne

TimBroddin:
## What

Adds hand-verified latitude/longitude for every Plopsaland **De Panne**
attraction and restaurant.

## Why

The Plopsa middleware feed (`/api/points-of-interest`) exposes only *pixel*
coordinates on the park-map image, not lat/lng. Plopsaland Deutschland has an
affine pixel→geo transform, but **De Panne doesn't** — so its attractions and
restaurants shipped with no `location`.

## How

- `PlopsaBase` gains `poiLocations` + `lookupPoiLocation()`, keyed by POI title
  with normalized matching (case-insensitive + curly/straight apostrophe
  folding). This mirrors the existing Enchanted Parks `attractionLocations`
  pattern.
- `buildEntityList()` now prefers this snapshot for both `ATTRACTION` and
  `RESTAURANT`, falling back to `mapCoordinates()` (Deutschland's pixel
  transform) when there's no snapshot entry — so Deutschland is unchanged.
- The De Panne constructor loads
  `locations/plopsaland-de-panne.json`.
- Coordinates were seeded from Google Places + Apple Maps (name-verified
  matches only) and finished by hand on a satellite map. Provenance and refresh
  instructions are in `locations/README.md`.

## Coverage

**86 of 87 POIs** located — all 34 restaurants, 52/53 attractions.
*Merry-go-Round* is intentionally absent (it is not shown on the park map).

## Testing

- 3 new unit tests for the lookup (known-ride resolution, apostrophe/case
  folding, unknown-title miss). Full suite green (1252 tests).
- Verified **end-to-end against the live API**: `getEntities()` builds 101
  entities with 52/53 attractions and 34/34 restaurants receiving coordinates;
  `getLiveData()` and `getSchedules()` pass.

## Notes

- No dependency, config, or behavior changes for any other park.
- Snapshot can be refreshed with the tooling linked in `locations/README.md`.

### DIFF
--- a/src/parks/plopsa/__tests__/plopsa.test.ts
+++ b/src/parks/plopsa/__tests__/plopsa.test.ts
@@ -8,7 +8,7 @@
  * disagrees between instances.
  */
 import {describe, test, expect} from 'vitest';
-import {plopsaDecideOperating} from '../plopsa.js';
+import {plopsaDecideOperating, Plopsaland} from '../plopsa.js';
 
 describe('plopsaDecideOperating', () => {
   test('park closed → ride always CLOSED regardless of other inputs', () => {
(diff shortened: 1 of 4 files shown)


cubehouse (changes requested):
Solid piece of work — the data provenance write-up in `locations/README.md` is genuinely better documentation than most of what's in this repo already, and mirroring the Enchanted Parks lookup pattern is the right call. Two things worth fixing before merge, one worth a quick check.

## 1. Null/missing `title` will crash the whole De Panne entity build

```ts
private normalizePoiTitle(title: string): string {
  return title.replace(/[‘’]/g, "'").toLowerCase();
}
```

`lookupPoiLocation()` calls this unguarded, and both call sites (`plopsa.ts:382`, `396`) pass `item.title` straight from the parsed HTTP JSON. `PlopsaContainsItem.title: string` is a compile-time annotation only — nothing validates it at runtime (`fetchPOI()` doesn't set `validateResponse`). If the middleware ever returns a `contains[]` item with `title` null or missing (placeholder row, malformed record), `title.replace(...)` throws, and there's no try/catch anywhere in `buildEntityList()` or in `destination.ts`'s `getEntities()` — so one bad record fails entity generation for the *entire* destination, not just that POI. This is De Panne-only; Deutschland is safe because `lookupPoiLocation` short-circuits on `!this.poiLocations` before ever touching `title`.

Five other parks in this repo already guard exactly this on their own feeds (`nigloland.ts`, `futuroscope.ts`, `qiddiyacity.ts`, `liseberg.ts`, `universalbeijing.ts` all check for falsy title). One-line fix: `if (!title) return '';` at the top of `normalizePoiTitle`, or an early return in `lookupPoiLocation`.

## 2. The new tests don't cover the actual wiring

All 3 tests call `Probe.lookup()` directly, which exercises `lookupPoiLocation()` in isolation. The two lines that actually wire the feature in — `plopsa.ts:382` and `396` — are never executed by any test, so nothing asserts an entity's `.location` ends up populated from the snapshot, or that the snapshot takes precedence over `poiCoords` (the whole point of the PR). `seaworld.test.ts:114-138` and `oceanpark.test.ts:362-383` both test `buildEntityList()` end-to-end by reassigning the instance's raw `@http`-decorated fetch method — same technique would drop in here at roughly the cost of the 3 tests already written, and it'd also catch #1 above if you add a null-title fixture case.

## Worth a quick look, not blocking

`"The Big & Betsy Farm"` and `"The Farmhouse at Big & Betsy"` share the exact same coordinates (51.0828926, 2.599892) in the snapshot. Plausibly two adjacent sub-features of the same themed area, but worth a 10-second confirmation it's not a copy-paste in the pinning tool rather than two independently-verified pins.

---

Root-cause reasoning and data provenance are both good. It's the unguarded external-data path and the test gap around it I'd want closed first.

TimBroddin:
Thanks for the review! Good catches!

- Hardened the De Panne POI path so null/missing `title` values no longer crash entity generation; titleless POI children are skipped instead of failing the whole destination.
- Reworked the location tests to exercise `buildEntityList()` end-to-end with mocked fetch methods, including snapshot population, snapshot precedence over mapped POI coords, curly apostrophe matching, and a null/missing-title fixture.
- Rechecked Big & Betsy and updated `The Farmhouse at Big & Betsy` from the corrected manual pin in `.plopsa-pinning/output.json`, so it no longer shares coordinates with `The Big & Betsy Farm`.

Validation: `npm run build`, `npm test`.

Also, credit where due: a lot of this was helped along by my good friends Fable & Codex 😅

PR #241, opened 2026-07-07
fix(oceanpark): rebuild on website scraping — mobile app API suspended

cubehouse:
## Summary
- Ocean Park's mobile app (and its API at sop.oceanpark.com.hk) has been suspended — every endpoint now returns a permanent 502 from Ocean Park's own gateway, which is why entities/live-data/schedules had been stale for about a week.
- Rebuilt the destination against the public website (www.oceanpark.com.hk), which server-renders the same data — including live per-attraction wait times — as an embedded React Server Component JSON payload. No auth required.
- Entity IDs changed: attractions are now keyed on the website's own node UUID, restaurants/shows on a URL/title slug, since the app's numeric IDs don't exist in this data source at all. Coordinates still come from map.oceanpark.com.hk, now joined by URL slug instead of the old numeric id.
- Known gaps vs. the old implementation (the website doesn't expose these): FastPass/paid-return-time flags, pregnancy/wet-ride warnings, per-attraction "Summit closes early" special hours, and shopping/food-kiosk locations (no stable id in the source data).

## Test plan
- [x] `npm run build` — clean typecheck
- [x] `npm test` — full suite green (1292 tests, including 43 new Ocean Park unit tests covering the RSC-JSON extractor, slug/queue/hour-range parsers, and the entity/live-data/schedule builders)
- [x] `npm run dev -- oceanparkhongkong` — live run against the real website: 71 entities (43 attractions, 8 restaurants, 18 shows), 61 live data entries (32 with real wait times), 56-day schedule

## Follow-up (blocking, needs Jamie)
Entity IDs changed for this destination. Per the migration workflow, this needs `npm run migrate -- oceanparkhongkong` run and confirmed before merging/enabling, plus `npm run migrate:audit` for the DESTINATION externalId.

### DIFF
--- a/TODO.MD
+++ b/TODO.MD
@@ -27,7 +27,7 @@
 - Universal Studios Singapore (USS)
 - Lotte World
 - Chimelong: Guangzhou + Zhuhai (2 destinations)
-- Ocean Park Hong Kong: Auth token, coordinate affine transform, showtimes, park schedule
+- Ocean Park Hong Kong: Website RSC-JSON scraping (mobile app API was suspended), coordinate affine transform, showtimes, park schedule
 
 **Totals: ~119 destinations, 1068 tests, 45 test files**
 
--- a/src/parks/oceanpark/__tests__/oceanpark.test.ts
+++ b/src/parks/oceanpark/__tests__/oceanpark.test.ts
@@ -1,0 +1,453 @@
+/**
+ * Unit tests for Ocean Park Hong Kong.
+ *
+ * The mobile app's API (sop.oceanpark.com.hk) was suspended along with the
+ * app; this park now scrapes the public website's server-rendered pages.
+ * These tests exercise the pure parsing helpers and the entity/live-data/
+ * schedule builders offline, without hitting the real site. Full integration
+ * is exercised via `npm run dev -- oceanparkhongkong`.
+ */
+import {describe, test, expect, vi, beforeEach, afterEach} from 'vitest';
+import {
+  OceanParkHongKong,
+  extractRscArray,
+  findMatchingBracket,
+  slugFromUrl,
+  slugify,
+  parseQueueMinutes,
+  parseHourRange,
+  computeAffineTransform,
+} from '../oceanpark.js';
+
+const TZ = 'Asia/Hong_Kong';
+
+// ── extractRscArray / findMatchingBracket ────────────────────────────────────
+
+describe('findMatchingBracket', () => {
+  test('finds the matching close bracket for a simple array', () => {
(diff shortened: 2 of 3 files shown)


cubehouse:
Ran a 9-lens multi-agent adversarial review against this diff (correctness, security, reliability/scraping-fragility, data-integrity, test-coverage, conventions, efficiency, timezone/dates, error-handling) — each finding independently verified by 3 skeptics before counting. 21 confirmed (2 critical, 4 high, 8 medium, 7 low), 5 refuted.

All confirmed findings fixed in the latest commit except one deliberate scope call:

**Fixed:**
- Malformed attraction/restaurant items (missing nodeUrl/nodeId) no longer crash the whole entity list — dropped and logged individually.
- `slugify()` collisions (e.g. "Whiskers & Friends" vs "Whiskers, Friends" both slugifying to the same id) no longer silently merge two shows — resolved deterministically with logging, shared between entity-list and live-data builders.
- Quadratic-time scan risk in the RSC bracket-parser on malformed/truncated HTML — capped input size and scan attempts.
- Silent empty-output on a marker miss now logs a warning, matching the existing coordinate-fetch pattern.
- Overnight closing times (e.g. "6:00 pm - 12:30 am") now roll to the next calendar day instead of producing closingTime < openingTime.
- Independent upstream sources (attractions/dining/shows/schedule/coordinates) each fail in isolation now — one flaky endpoint degrades that category instead of losing the whole build.
- `buildSchedules` uses `Promise.allSettled` across its 60 per-date requests so one bad date drops one day, not the whole 60-day schedule.
- Replaced host-timezone-dependent date stepping with pure calendar-string arithmetic (removes a DST-adjacent off-by-one-day risk on non-UTC hosts).
- Plus test-coverage gaps: MAXIMUM_HEIGHT branch, overnight hour ranges, coordinate fallback for restaurants/shows, empty `attractionTypes`, escaped-backslash edge cases — 39 new tests total (43 → 82).

**Deliberately not fixed:**
- Sharing the RSC-extraction utility with Futuroscope's separate, weaker implementation — real improvement, but a cross-park refactor that's better scoped as its own follow-up rather than riding on this bug-fix PR.
- Grouping same-titled shows by venue (a refuted finding — plausible but speculative; would need schedule data this rewrite doesn't currently fetch, and the daily-schedule feed's actual `locations` field wasn't wired up for this).

Verified: `npm run build` clean, full suite 1331/1331 (was 1292), live run against the real site — 71 entities, schedule/live-data both correct (park is closed right now at 22:05 HKT, correctly reporting CLOSED with no fabricated wait times).

PR #243, opened 2026-07-07
fix(efteling): align single-rider lifecycle with STANDBY

BeyondVertical:
## Summary

The Efteling adapter derived `queue.SINGLE_RIDER` **only** from the single-rider
entry's `WaitingTime` and ignored that entry's own `State` field. As a result a
**closed** single rider and an **open** single rider **without a posted wait**
were emitted identically as `SINGLE_RIDER: { waitTime: null }` — consumers could
not tell them apart, even though the source feed distinguishes them.

This PR implements the minimal, adapter-only fix (no schema change): read the
single-rider entry's own `State` and **omit** the `SINGLE_RIDER` queue when the
single rider isn't operating.

## The source feed carries the distinction

Efteling's WIS feed lists each single rider as its own `AttractionInfo` entry
(`…singlerider`) with its **own `State`** field (`open` / `gesloten`),
independent of the parent attraction. Two ambiguous cases, verbatim:

```json
{"Id": "baron1898singlerider",    "Name": "Baron 1898 Single-rider",    "Type": "Attracties", "State": "gesloten", "Empire": "Ruigrijk"}
{"Id": "dansemacabresinglerider", "Name": "Danse Macabre Single Rider", "Type": "Attracties", "State": "open",     "Empire": "Anderrijk"}
```

Neither carries a `WaitingTime`; the **only** thing separating closed from open
here is `State`.

## Root cause

The adapter identifies single-rider entries correctly (via POI metadata,
`alternatetype === 'singlerider'` → parent id), but when reading their live data
it took **only `WaitingTime`** and never consulted the single-rider entry's own
`State`. The `SINGLE_RIDER` queue was then built from the **parent attraction's**
status. Because both parents were `open` and both single-rider entries lacked
`WaitingTime`, both collapsed to `waitTime: null` — the `gesloten` vs `open`
signal the feed provides was dropped.

## The fix

- **First pass** now captures each single-rider entry's own `State` alongside
  its wait time (`{ status, waitTime }` per parent), instead of just the wait time.
- **Second pass** drives the `SINGLE_RIDER` queue from the single rider's **own**
  status (not the parent's) and **omits** the queue entirely when it isn't
  operating.

```ts
// First pass
const singleRiderData = new Map<string, {status: string; waitTime: number | null}>();
...
singleRiderData.set(parentId, {
  status: this.mapState(entry.State),
  waitTime: Number.isFinite(waitTime) ? waitTime : null,
});

// Second pass
const sr = singleRiderData.get(entityId);
if (sr && sr.status === 'OPERATING') {
  ld.queue.SINGLE_RIDER = { waitTime: sr.waitTime };
}
```

### Behaviour after the fix

| Single-rider `State` / `WaitingTime` | Emitted `queue.SINGLE_RIDER` | Meaning |
|---|---|---|
| `gesloten`, no wait | *(queue omitted)* | closed — like any attraction without a single-rider queue |
| `open`, no wait | `{ waitTime: null }` | open, no posted wait |
| `open`, `N` | `{ waitTime: N }` | open, N-minute wait |

A closed single rider now simply has no queue. Consumers that track a
"last seen" value treat the disappearing queue as closed.

## Verification (live WIS feed, 2026-07-07)

Ran the dev harness (`npm run dev -- efteling`, 4/4 passing) and cross-referenced
the raw WIS `State` with the adapter's emitted queues:

| WIS single-rider entry | `State` | `WaitingTime` | Emitted `SINGLE_RIDER` |
|---|---|---|---|
| `baron1898singlerider` | `gesloten` | — | **omitted** ✓ |
| `devliegendehollandersinglerider` | `gesloten` | — | **omitted** ✓ |
| `dansemacabresinglerider` | `open` | — | `{ waitTime: null }` ✓ |
| `jorisendedraaksinglerider` | `open` | 20 | `{ waitTime: 20 }` ✓ |
| `maxmoritzsinglerider` | `open` | 5 | `{ waitTime: 5 }` ✓ |
| `pythonsinglerider` | `open` | 0 | `{ waitTime: 0 }` ✓ |
| `symbolicasinglerider` | `open` | 5 | `{ waitTime: 5 }` ✓ |

Both `gesloten` single riders are now omitted (no phantom `{ waitTime: null }`),
while the open-without-wait case (`dansemacabre`) keeps `{ waitTime: null }` —
exactly the distinction the report asked for.

## Commits

1. **fix(efteling): distinguish closed vs open single-rider queues** — the fix above.
2. **refactor(efteling): drop unused datetime imports** — `formatUTC`,
   `parseTimeInTimezone`, `formatInTimezone`, `isBefore` were imported but never used.
3. **refactor(efteling): use `Number.isFinite` for STANDBY wait time** — aligns the
   STANDBY guard with CLAUDE.md's numeric-validation rule and the single-rider code.
   Behaviour is unchanged (`parseInt` already yields `NaN` for empty/invalid input).

### DIFF
--- a/src/parks/efteling/__tests__/singlerider.test.ts
+++ b/src/parks/efteling/__tests__/singlerider.test.ts
@@ -1,0 +1,107 @@
+/**
+ * Single-rider queue lifecycle. The WIS feed lists each single rider as its own
+ * entry with its OWN State, independent of the parent. These pin:
+ *
+ *   - the SINGLE_RIDER key stays present for every capable ride (no flapping);
+ *   - waitTime holds a number only when the rider's own state is OPERATING,
+ *     else null;
+ *   - a closed/down rider under an OPEN parent keeps the key, and a ride with
+ *     no single-rider alternate never grows the key.
+ *
+ * Fetch methods are stubbed; full integration runs via `npm run dev -- efteling`.
+ */
+
+import {describe, test, expect} from 'vitest';
+import {Efteling} from '../efteling.js';
+
+/** A single-rider-capable attraction POI: parent id + its single-rider alternate. */
+function srPoi(id: string) {
+  return {
+    fields: {
+      id,
+      category: 'attraction',
+      latlon: '51.65,5.05',
+      alternatetype: 'singlerider',
+      alternateid: `${id}singlerider`,
+    },
+  };
+}
+
+/** A plain attraction POI with no single-rider alternate. */
+function plainPoi(id: string) {
+  return {fields: {id, category: 'attraction', latlon: '51.65,5.05'}};
+}
+
+const POI_HITS = [
+  srPoi('python'), // open single rider with a posted wait
+  srPoi('joris'), // open single rider, no posted wait
+  srPoi('baron'), // single rider gesloten under an OPEN parent
+  srPoi('max'), // single rider mechanical fault (storing -> DOWN)
+  plainPoi('carnaval'), // no single-rider alternate at all
+];
+
+const WAIT_TIMES = [
+  // Parents — all open, so the single rider's own state is the only variable.
+  {Id: 'python', Type: 'Attracties', State: 'open', WaitingTime: 25},
+  {Id: 'joris', Type: 'Attracties', State: 'open', WaitingTime: 15},
+  {Id: 'baron', Type: 'Attracties', State: 'open', WaitingTime: 20},
+  {Id: 'max', Type: 'Attracties', State: 'open', WaitingTime: 10},
+  {Id: 'carnaval', Type: 'Attracties', State: 'open', WaitingTime: 30},
+  // Single-rider rows — their Id is the alternate, not a POI id.
+  {Id: 'pythonsinglerider', Type: 'Attracties', State: 'open', WaitingTime: 5},
+  {Id: 'jorissinglerider', Type: 'Attracties', State: 'open'},
+  {Id: 'baronsinglerider', Type: 'Attracties', State: 'gesloten'},
+  {Id: 'maxsinglerider', Type: 'Attracties', State: 'storing'},
+];
+
+function mockedEfteling(poiHits: any[], waitTimes: any[]): Efteling {
+  const park = new Efteling();
+  (park as any).getPOIData = async () => poiHits;
+  (park as any).getWaitTimes = async () => waitTimes;
+  return park;
+}
+
+async function liveById(poiHits = POI_HITS, waitTimes = WAIT_TIMES) {
+  const live = await (mockedEfteling(poiHits, waitTimes) as any).buildLiveData();
+  return (id: string) => live.find((l: any) => l.id === id);
+}
+
+describe('Efteling buildLiveData — single rider queue', () => {
+  test('surfaces the posted single-rider wait when the single rider is open', async () => {
+    const get = await liveById();
+    expect(get('python').queue.SINGLE_RIDER).toEqual({waitTime: 5});
+  });
+
+  test('keeps the key present with waitTime: null when open without a posted wait', async () => {
+    const get = await liveById();
+    const joris = get('joris');
+    expect(joris.queue).toHaveProperty('SINGLE_RIDER');
+    expect(joris.queue.SINGLE_RIDER).toEqual({waitTime: null});
+  });
+
+  test('keeps the key present when the single rider is closed — no flapping', async () => {
+    const get = await liveById();
+    const baron = get('baron');
+    // Parent is open; only the single rider is gesloten. The key must NOT
+    // disappear across polls.
+    expect(baron.queue).toHaveProperty('SINGLE_RIDER');
+    expect(baron.queue.SINGLE_RIDER).toEqual({waitTime: null});
+    // The parent's own STANDBY queue is unaffected.
+    expect(baron.queue.STANDBY).toEqual({waitTime: 20});
+  });
+
+  test('a single-rider mechanical fault (storing/DOWN) keeps the key, not omitted', async () => {
+    const get = await liveById();
+    const max = get('max');
+    // DOWN/REFURBISHMENT must not be conflated with CLOSED by dropping the key.
+    expect(max.queue).toHaveProperty('SINGLE_RIDER');
+    expect(max.queue.SINGLE_RIDER).toEqual({waitTime: null});
+  });
+
+  test('never grows a SINGLE_RIDER key for a ride with no single-rider alternate', async () => {
+    const get = await liveById();
+    const carnaval = get('carnaval');
+    expect(carnaval.queue).not.toHaveProperty('SINGLE_RIDER');
+    expect(carnaval.queue.STANDBY).toEqual({waitTime: 30});
+  });
+});
--- a/src/parks/efteling/efteling.ts
+++ b/src/parks/efteling/efteling.ts
@@ -407,15 +407,20 @@ export class Efteling extends Destination {
     const waitTimes = await this.getWaitTimes();
 
     // First pass: collect single rider data
-    // WIS returns separate entries for single rider queues using the alternate ID
-    const singleRiderData = new Map<string, number | null>();
+    // WIS returns separate entries for single rider queues using the alternate ID.
+    // Each single rider entry carries its OWN State (open/gesloten), independent of
+    // the parent attraction, so capture it alongside the wait time.
+    const singleRiderData = new Map<string, {status: string; waitTime: number | null}>();
     for (const entry of waitTimes) {
       if (!entry.Id) continue;
       // If this ID is NOT a known POI but IS a single rider alternate ID
       if (!poiData.has(entry.Id) && singleRiderMap.has(entry.Id)) {
         const parentId = singleRiderMap.get(entry.Id)!;
-        const waitTime = entry.WaitingTime !== undefined ? parseInt(String(entry.WaitingTime), 10) : null;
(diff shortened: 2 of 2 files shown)


cubehouse (changes requested):
Nice catch on the root cause — the WIS feed really does carry the single-rider's own `State` independently of the parent, and the before/after verification table against the live feed is a good way to prove it. Ran this through a deeper multi-angle review and found a couple of things worth fixing before merge, plus a gap in test coverage.

## 1. This reintroduces a flapping bug DLP already had to fix

`disneylandparis.ts:961-984,1051-1053` keeps a `CacheLib`-persisted `seenSR` set specifically so `SINGLE_RIDER` never disappears once observed — only `waitTime` toggles between a number and `null`. The comment there is explicit: *"Consumers expect these queues to stay present so they can render `wait: null` rather than disappearing."*

This PR does the opposite: `ld.queue.SINGLE_RIDER` is now omitted entirely whenever the single-rider entry's own state isn't `OPERATING`, with no persistence (`efteling.ts` has zero `CacheLib` usage). Your own verification table shows 2 of 7 single riders were `gesloten` at test time — that's not a rare edge case, that's lunch breaks and routine short stoppages. Every one of those will now flap the `SINGLE_RIDER` key in and out of existence across consecutive polls, which is exactly the transition DLP's author already paid to prevent elsewhere in this codebase.

Worth either porting DLP's `seenSR` pattern here, or explicitly deciding Efteling doesn't need it and saying why (e.g. if you've confirmed the collector treats an absent key the same as `waitTime: null` — see #3 below).

## 2. DOWN/REFURBISHMENT gets silently conflated with CLOSED

```ts
const sr = singleRiderData.get(entityId);
if (sr && sr.status === 'OPERATING') {
  ld.queue.SINGLE_RIDER = { waitTime: sr.waitTime };
}
```

`STANDBY` a few lines up keeps its key present (`waitTime: undefined`) when the parent is `DOWN`/`REFURBISHMENT`. `SINGLE_RIDER`'s new gate collapses `DOWN`/`REFURBISHMENT` into the same "omit entirely" bucket as `CLOSED`/`gesloten` — since `mapState()` is the same generic function applied to both the parent and the single-rider's own `State`, this is a reachable state, not a hypothetical. A single-rider-specific mechanical fault (state `storing`) now looks identical to a full-day closure. Given the PR's whole point is disambiguating "closed" from "open," it'd be good to either mirror STANDBY's presence rule here (key stays, `waitTime: undefined`/`null` for down/refurb) or state explicitly that down/refurb collapsing into "closed" is intentional.

## 3. No test — and there's a direct in-repo template for one

`src/parks/efteling/` has no `__tests__/` directory, but this isn't a "parks aren't unit tested" repo-wide thing — `dlp/`, `universal/`, `plopsa/`, `seaworld/`, `europapark/`, `tdr/`, `enchantedparks/` all have one. `dlp/__tests__/visibility.test.ts` is close to a drop-in template: instantiate the class, `vi.spyOn` the private fetch methods with synthetic fixtures, assert on output, no HTTP mocking needed, runs in milliseconds.

For Efteling specifically, mocking `getPOIData('en')` (single-rider POI + parent) and `getWaitTimes()` (gesloten / open-no-wait / open-with-wait single-rider rows) and asserting on `ld.queue.SINGLE_RIDER` would cover exactly the three cases already in your verification table — and pin the behavior so a future merge/revert can't silently reintroduce the ambiguity this PR fixes. The live-feed table is convincing for today, but it's prose, not code — nothing in CI would catch a regression here.

## Smaller note

The "disappearing queue = closed" assumption is checked against the raw WIS feed, but not against how the (closed-source) collector actually interprets a missing key vs `waitTime: null`. Given DLP's comment implies a real consumer broke on this exact kind of omission before, worth a quick sanity check there too — independent of what you decide on #1.

---

Nothing here is about the root-cause diagnosis, which is solid. It's the "omit vs. always-present" design choice and test coverage I'd want settled before this ships.

BeyondVertical:
Thanks — really useful review. Following it properly turned up a schema-level issue that I think is the actual crux, so let me put it plainly. I've pushed a branch that addresses all four points without a schema change, but I want to flag upfront that on its own it gives up the exact distinction this PR is named for. Here's why.

## The problem

The WIS feed gives every single rider its own `State` (`open` / `gesloten` / `storing` / `inonderhoud`) plus an optional `WaitingTime`, independent of the parent. So an SR entrance has four meaningful conditions:

1. open, with a posted wait
2. open, no posted wait
3. closed (`gesloten`) — often while the ride itself keeps running
4. down / refurbishment

But `SINGLE_RIDER` is typed `{ waitTime: number | null }` — only three distinguishable shapes: key **absent**, `{waitTime: null}`, `{waitTime: N}`. The one shape that separated **closed** from **open-no-wait** in the original version was "key absent", which is precisely the flapping you flagged (`gesloten` toggles all day). Remove it to stop the flapping and only two shapes remain for those two conditions.

**So no-flapping and "closed ≠ open-no-wait" can't both hold under this schema.** The lost case is the one the PR targeted: parent `OPERATING` + SR `gesloten` emits `{waitTime: null}`, identical to open-with-no-wait. `status` only reflects the parent, so it can't cover the SR line while the ride runs.

The other queue types already solve this — `RETURN_TIME` carries a `state`, `BOARDING_GROUP` an `allocationStatus`; only the wait-based queues are stateless. So it's an existing inconsistency, not a new feature.

## Suggested fix — a per-queue `state`

```ts
SINGLE_RIDER?: {
  waitTime: number | null;
  state?: LiveStatusType;   // OPERATING | CLOSED | DOWN | REFURBISHMENT
};
```

```ts
ld.queue.SINGLE_RIDER = {
  waitTime: (sr && sr.status === 'OPERATING' && sr.waitTime !== null) ? sr.waitTime : null,
  state: (sr?.status ?? 'CLOSED') as LiveStatusType,
};
```

Key stays present (no flapping) **and** all four conditions are recoverable — including DOWN vs. REFURBISHMENT, which resolves #2 properly rather than just documenting the collapse. It matches the queues that already carry state and generalizes to other parks. Cost is a `typelib` change + version bump — your call, which is why I didn't just do it. Happy to put up the typelib PR and wire this side to match.

## What's on the branch now (no schema change)

Addresses all four points, but accepts that conditions 2–4 merge:

- **#1** key stays present for every SR-capable ride; capability comes from the stable POI feed, so no `seenSR`/`CacheLib` set is needed (Efteling's signal never goes silent like DLP's wait feed)
- **#2** mirrors STANDBY; down/refurb/closed/open-no-wait all → `{waitTime: null}`, documented as intentional
- **#3** new `efteling/__tests__/singlerider.test.ts`, following `dlp/visibility.test.ts`
- **#4** key never dropped → no absent-vs-`null` ambiguity

Tradeoff: it drops the distinction the PR was named for, so it'd want a retitle (e.g. "align single-rider lifecycle with STANDBY") — or closing, if that's too thin to stand alone.

<sub>(A no-schema shortcut — `waitTime: 0` for open-no-wait, `null` for closed — separates them too, but implies a "0-min walk-on" we haven't confirmed. Noting it only for completeness; I'd lean against it.)</sub>

## Over to you

I'd suggest the per-queue `state` — it's the clean fix and resolves #2 properly. But the branch is a real fallback: I can retitle it, or close it, whatever you prefer. Just let me know.

cubehouse:
Good catch on the schema-level tension — that's the right way to have surfaced it rather than picking a side quietly.

Decision: no `typelib` change for now. Per-queue `state` on `SINGLE_RIDER` (and generalizing it to `STANDBY`) is a real improvement but it's a v2-scope change, not something to fold into this fix. Go with the fallback on the current branch — key stays present, down/refurb/closed/open-no-wait all collapse to `{waitTime: null}`, matches STANDBY.

One ask before merge: `sr.status` (`efteling.ts` around line 470) already carries the exact OPERATING/CLOSED/DOWN/REFURBISHMENT signal your proposed `state` field would need — it's computed, just not emitted. Rather than that fact living only in this PR conversation, drop a comment at the `ld.queue.SINGLE_RIDER = {...}` assignment noting that `sr.status` is ready to wire into a future `state` field once `typelib` supports it, so whoever picks up v2 doesn't have to re-derive this. No behavior change, just a marker.

Also: go ahead and retitle this PR (or the final commit) to reflect what it now actually does — something like "align single-rider lifecycle with STANDBY" — since the closed-vs-open-no-wait distinction it was originally named for is the part being deferred.

BeyondVertical:
All three sorted — thanks for the clear call on deferring `state` to v2.

- **Fallback kept** — the key stays present for every single-rider-capable ride, and `down`/`refurb`/`closed`/open-no-wait all collapse to `{waitTime: null}`, matching STANDBY. No `typelib` change.
- **v2 marker added** (`e592206a`) — a comment sits right at the `ld.queue.SINGLE_RIDER = {…}` assignment noting that `sr.status` already carries the OPERATING/CLOSED/DOWN/REFURBISHMENT signal, ready to wire into a `state` field once typelib gains one, so v2 doesn't have to re-derive it.
- **Retitled** to "align single-rider lifecycle with STANDBY".

Also brought the branch up to date with `main` (clean merge — the new `efteling.test.ts` from the venue-hours PR and the `singlerider.test.ts` here sit side by side, 7 tests green). Ready for another look whenever.

cubehouse (approved):
Fix commits address all three points raised: the single-rider key now stays present (derived from POI capability rather than the volatile wait feed, verified live — no more flapping), DOWN/REFURBISHMENT/CLOSED collapse to `waitTime: null` mirroring STANDBY, and `singlerider.test.ts` covers the lifecycle. Verified against the live WIS feed just now.

PR #245, opened 2026-07-10
feat(parcsreunidos): move Stay-App auth to an external token service

cubehouse:
## What

`STAYAPP_AUTHTOKEN` was a static bearer that silently goes stale (no refresh
logic anywhere — see the current 401s across Madrid/Mirabilandia/Movie Park
Germany/Bobbejaanland). Replaces it with an external-token-service pattern:

```
STAYAPP_TOKENURL=https://api.themeparks.wiki/v1/docs/id/stayapp_token
```

An out-of-process job is responsible for keeping that doc's `accessToken`
rolling (the app itself gets its bearer via an OAuth2 password grant against
a Keycloak realm using a fixed, app-embedded service-account credential
shared across every Parcs Reunidos park — that credential deliberately isn't
stored in this repo).

## How

- **`src/tokenService.ts`** (new) — `fetchExternalToken()`, extracted out of
  `GentingSkyworlds.getAccessToken()` so both parks share one implementation.
  Cached by `tokenUrl` rather than by calling class, so multiple destinations
  pointed at the same token endpoint share one fetch instead of each polling
  independently.
- **`gentingskyworlds.ts`** — `getAccessToken()` now delegates to the shared
  helper. No behavior change (verified: same log format, same graceful
  degradation on failure).
- **`parcsreunidos.ts`** — `authToken` (static) replaced with
  `tokenUrl`/`tokenAuth`/`tokenAuthHeader` (external-service pattern),
  wired into `injectHeaders()`.

## A bug I caught via live testing, not just unit tests

Unlike Genting's optional VQ bearer, every Stay-App endpoint requires this
header. My first pass unconditionally set `Authorization: Bearer ${token}`,
so an empty token produced `Bearer ` (trailing space, no value). Tested live
against the real API — Stay-App rejects that with
`AUTHORIZATION_HEADER_BAD_FORMED`, which is a *worse* failure mode than
today's `AUTHENTICATION_FAILED`. Fixed: `injectHeaders()` now omits
`Authorization` entirely when no token is available, reproducing the
original (expected) failure mode instead of a new one.

## Testing

- 9 new unit tests for `tokenService.ts` (empty tokenUrl, custom auth
  header, default header, cache hit, cross-destination cache sharing, HTTP
  failure not cached, missing-accessToken response).
- 4 new unit tests for `ParcsReunidosDestination.injectHeaders` (token
  injection, custom auth header, Authorization omitted when unconfigured,
  cross-park cache sharing).
- Full suite: 1395/1395 passing.
- Live-verified against the real Genting token endpoint (401, degrades
  gracefully, log format unchanged from pre-refactor).
- Live-verified against the real Stay-App API with no token configured
  (correctly reproduces `AUTHENTICATION_FAILED`, not the header-format bug
  above).

## Deployment note

This alone doesn't fix the live 401s — `STAYAPP_TOKENURL` needs an actual
out-of-process refresher pushing to `docs/id/stayapp_token` before Parcs
Reunidos parks resume. That's tracked separately; this PR is the
`parksapi`-side half.

### DIFF
--- a/src/__tests__/frameworkContract.test.ts
+++ b/src/__tests__/frameworkContract.test.ts
@@ -97,15 +97,15 @@ describe('@destinationController auto-applies @config', () => {
   });
 
   test('config prefix resolution works for registered parks', async () => {
-    process.env.STAYAPP_AUTHTOKEN = 'test-token-123';
+    process.env.STAYAPP_TOKENURL = 'https://token.example/shared-stayapp-doc';
 
     const entry = await getDestinationById('movieparkgermany');
     const park = new entry!.DestinationClass();
 
     // Should resolve via STAYAPP prefix (shared)
-    expect((park as any).authToken).toBe('test-token-123');
+    expect((park as any).tokenUrl).toBe('https://token.example/shared-stayapp-doc');
 
-    delete process.env.STAYAPP_AUTHTOKEN;
+    delete process.env.STAYAPP_TOKENURL;
   });
 });
 
--- a/src/__tests__/tokenService.test.ts
+++ b/src/__tests__/tokenService.test.ts
@@ -20,13 +20,13 @@ describe('fetchExternalToken', () => {
   });
 
   test('returns empty string when tokenUrl is unset', async () => {
-    const token = await fetchExternalToken({tokenUrl: ''});
+    const token = await fetchExternalToken({tokenUrl: '', cacheKeyPrefix: 'TestPark'});
     expect(token).toBe('');
     expect(fetchMock).not.toHaveBeenCalled();
   });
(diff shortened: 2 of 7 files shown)


cubehouse:
Closing for now — the design here is solving the wrong problem. Turns out the actual auth mechanism these apps use for the attraction/establishment API is a static, long-lived token baked into the app binary (shared across parks, confirmed via live traffic from two different park apps), not the OAuth2/Keycloak flow this PR built a whole external-token-service pipeline around — that flow backs an unrelated feature (the survey/questionnaire endpoint). Revisiting with the corrected understanding.

PR #246, opened 2026-07-10
feat(plopsa): surface live show times & keep them when the calendar fails

BeyondVertical:
## Summary

Shows and meet-and-greets at both Plopsa parks (De Panne and Deutschland) were
only exposed as **entities** — they never appeared in **live data**, so consumers
saw no status and no today's performance times for them. Their **schedules were
also silently dropped** whenever the park-calendar fetch failed, even though the
show times come from a completely separate feed.

This PR fixes both, entirely within the shared `PlopsaBase` class (so both parks
benefit) and with no schema change:

1. **`fix`** — decouple the show schedules from the park-calendar fetch.
2. **`feat`** — emit each show/meet-and-greet in live data with today's
   performances as `showtimes` and an `OPERATING` / `CLOSED` status.
3. **`fix`** — guard the entertainments fetch so a failure there can never take
   down the ride live data or the park calendar (from review, see below).

## Problem 1 — a calendar failure discarded the show schedules

`buildSchedules` fetched the park calendar first and **early-returned `[]`** the
moment that fetch threw — *before* it built the per-show schedules from the
independent `entertainments` feed. A single transient calendar hiccup (or a
missing `*_CALENDARURL`) therefore took the park hours **and** all show times
down together.

Split into `buildParkSchedule()` (calendar feed) and `buildShowSchedules()`
(entertainments feed), combined at the end, so a failure in one never discards
the other.

## Problem 2 — shows/meet-and-greets had no live data

`buildLiveData` mapped **only** the wait-times feed, which lists attractions. It
now also reads the `entertainments` feed and emits a live entry per
show/meet-and-greet, via two exported, unit-tested pure helpers:

- **`plopsaBuildShowtimes(schedule, date, tz)`** — picks today's day entry, keeps
  only `open` slots, normalises HH:MM start times to ISO+offset via
  `constructDateTime` (DST-correct). The feed has no end time (point-in-time
  performances), so `endTime` is `null` unless the feed provides one.
- **`plopsaShowStatus(showtimes, nowMs)`** — `OPERATING` while a performance is
  still upcoming or ongoing, `CLOSED` once the day's last one has ended.

| Today's performances | `status` | `showtimes` |
|---|---|---|
| none listed for today | `CLOSED` | *(omitted)* |
| last one still upcoming/ongoing | `OPERATING` | today's slots |
| last one already passed | `CLOSED` | today's slots (kept for reference) |

### Consistency with the codebase

This is the **same rule the AttractionsIO parks use for shows** (see
`src/parks/attractionsio/attractionsiov1.ts`) — deliberately mirrored so the same
concept is expressed the same way: same `end = endTime ?? startTime` fallback,
same `Date.parse(end) > nowMs` predicate, same `type: 'Showtime'` slot label, same
"attach `showtimes` only when non-empty" rule.

## Problem 3 — an entertainments outage could take down unrelated data (from review)

Thanks to @cubehouse for catching this. Neither `getLiveData()` nor
`getSchedules()` wraps its `build*` method in a try/catch, so an unguarded
rejection propagates out and drops **everything** that method would have
returned.

- `buildLiveData`'s new `fetchEntertainments()` call had no guard, so a failing
  entertainments feed rejected the whole `Promise.all` and took the **ride
  wait-time live data** down with it — data this endpoint could not affect at all
  before this branch. Now gated with `.catch(() => null)`, matching
  `fetchTodayHours` right above it.
- `buildShowSchedules()` had the **mirror image** of the same defect: an
  entertainments failure rejected `buildSchedules()` and discarded the **park
  calendar** too. Now wrapped in the same try/catch `buildParkSchedule()` uses.

Both paths have regression tests, each verified to **fail without its guard**
(`Error: entertainments 500`) rather than passing vacuously.

## Merge with #249

`main` picked up #249 (*treat 0-wait as no signal, report DOWN not OPERATING*),
which rewrote the same `buildLiveData` ride-mapping block and renamed
`plopsaDecideOperating` → `plopsaDecideStatus`. Merged and resolved keeping
**both** sides:

- **#249's ride semantics** — `plopsaDecideStatus`, `rawWait` (only a strictly
  positive reading counts as a live signal), `DOWN` for park-open +
  `temporarily_closed` + no signal, and `STANDBY.waitTime = rawWait`.
- **This branch** — show/meet-and-greet live data, the `waitTimes ?? {}`
  relaxation, and the entertainments-fetch guards.

Verified live that both coexist correctly (see below).

## Verification (live feeds)

| Park | Harness | Attractions (live) | Shows (live) | Schedules |
|---|---|---|---|---|
| Plopsaland Deutschland | ✅ 4/4 | 47/47 — 46 OPERATING, **1 DOWN** | 17/17 — 14 with showtimes | 17 |
| Plopsaland De Panne | ✅ 4/4 | 53/53 — 47 OPERATING, **6 DOWN** (incl. SuperSplash) | 13/13 — 6 with showtimes | 10 |

#249's `DOWN` semantics and this branch's show live data work side by side.

Real emitted show data, spot-checked against the official app:

```
Plopsaland Deutschland (park 10:00–22:30)
  [OPERATING] Meet Holly                     16:50          ← still upcoming
  [CLOSED   ] Meet Vic & Halvar              13:25, 15:10   ← last one passed, times kept
  [CLOSED   ] Meet the Easter Bunny Happy    (none today)   ← seasonal, correct

Plopsaland De Panne (park 10:00–19:00)
  [OPERATING] Bumba Show                     13:00, 17:15
  [CLOSED   ] Maya Show                      12:00
  [CLOSED   ] Meet the Christmas Gnome       (none today)   ← seasonal, correct
```

Seasonal characters (Easter / Pumpkin / Christmas) correctly report `CLOSED` with
no performances in July; the "last performance passed → CLOSED, showtimes kept"
rule is visible on *Meet Vic & Halvar* vs *Meet Holly*.

### Tests

`src/parks/plopsa/__tests__/plopsa.test.ts` — **20/20 passing** (#249's rewritten
`plopsaDecideStatus` matrix retained, plus four new blocks):

- `plopsaBuildShowtimes` — date matching, non-`open` slot filtering, ISO+offset
  construction, `null` end times.
- `plopsaShowStatus` — empty → `CLOSED`; upcoming → `OPERATING`; last performance
  passed → `CLOSED`; start-time fallback when no end time.
- `buildSchedules` feed independence — show schedules survive a calendar failure;
  the park schedule survives an entertainments failure.
- `buildLiveData` — shows emit independently of the wait-times feed; a failing
  entertainments feed drops only the shows, never the ride live data; parades
  (not mapped as entities) are excluded.

## Deployment / configuration

No hardcoded URLs (per `CLAUDE.md`, these live in `.env`). Show times and the park
calendar only work when configured:

| Env var | Purpose |
|---|---|
| `PLOPSALANDDEUTSCHLAND_BASEURL` | Deutschland middleware base |
| `PLOPSALANDDEUTSCHLAND_CALENDARURL` | Deutschland opening-hours calendar |
| `PLOPSALAND_BASEURL` | De Panne middleware base |
| `PLOPSALAND_CALENDARURL` | De Panne opening-hours calendar |

`*_CALENDARURL` is the calendar endpoint **without** `start`/`end` —
`fetchCalendar` appends those itself. De Panne's `/en/plopsaland-de-panne/` slug
301-redirects to `/en/plopsaland-belgium/`; configure the redirect target directly.

## Out of scope

- **De Panne "Merry-go-Round" POI location** — left untouched. It is documented as
  *intentionally absent* in `locations/README.md` (not shown on the park map), and
  the map is not affine-clean enough to derive a hand-verified coordinate (median
  ~10 m, max ~106 m error across 86 control points).
- **Restaurant live data** — Plopsa has never emitted it; the POI feed exposes only
  a `temporarily_closed` flag, no hours. Unchanged here.

### DIFF
--- a/src/parks/plopsa/plopsa.ts
+++ b/src/parks/plopsa/plopsa.ts
@@ -519,6 +519,23 @@ class PlopsaBase extends Destination {
   // ── Schedules ─────────────────────────────────────────────────
 
   protected async buildSchedules(): Promise<EntitySchedule[]> {
+    const schedules: EntitySchedule[] = [];
+
+    // The park calendar and the show schedules come from separate feeds, so
+    // they are built independently: a failure fetching one must not discard
+    // the other. The park entry is only emitted when the calendar yields days.
+    const parkSchedule = await this.buildParkSchedule();
+    if (parkSchedule.length > 0) {
+      schedules.push({id: this.parkId, schedule: parkSchedule} as EntitySchedule);
+    }
+
+    schedules.push(...await this.buildShowSchedules());
+
+    return schedules;
+  }
+
+  /** Park operating hours for the next ~90 days from the calendar endpoint. */
+  private async buildParkSchedule(): Promise<EntitySchedule['schedule']> {
     const now = new Date();
     const startDate = formatDate(now, this.timezone);
     const endDate = formatDate(addDays(now, 90), this.timezone);
@@ -557,7 +574,11 @@ class PlopsaBase extends Destination {
       }
     }
 
-    // Also build show schedules from entertainments
+    return schedule;
+  }
+
+  /** Per-show operating schedules from the entertainments feed. */
+  private async buildShowSchedules(): Promise<EntitySchedule[]> {
     const entResp = await this.fetchEntertainments();
     const entData = (await entResp.json()) as PlopsaEntertainmentResponse;
     const showSchedules: EntitySchedule[] = [];
@@ -589,10 +610,7 @@ class PlopsaBase extends Destination {
       }
     }
 
-    return [
-      {id: this.parkId, schedule} as EntitySchedule,
-      ...showSchedules,
-    ];
+    return showSchedules;
   }
 }
 
--- a/src/parks/plopsa/plopsa.ts
+++ b/src/parks/plopsa/plopsa.ts
@@ -487,13 +487,13 @@ class PlopsaBase extends Destination {
       this.fetchWaitTimes(),
       this.fetchPOI(),
       this.fetchTodayHours(today).catch(() => null),
-      this.fetchEntertainments(),
+      this.fetchEntertainments().catch(() => null),
     ]);
 
     const waitTimes = (await waitResp.json()) as PlopsaWaitTimesResponse;
     const poiData = (await poiResp.json()) as PlopsaPOIResponse;
     const hoursData = hoursResp ? (await hoursResp.json()) as PlopsaTodayHours : null;
-    const entData = (await entResp.json()) as PlopsaEntertainmentResponse;
+    const entData = entResp ? (await entResp.json()) as PlopsaEntertainmentResponse : null;
 
     // Per-attraction temporarily-closed flag from POI.
     const closedById = new Map<string, boolean>();
@@ -565,10 +565,12 @@ class PlopsaBase extends Destination {
       } as unknown as LiveData;
     });
 
-    // Show / meet-and-greet live data. The entertainments feed is independent
-    // of the wait-times feed above, so these surface even when a ride feed
-    // hiccups. Today's performances are attached as showtimes and the show is
-    // CLOSED once the last one has passed.
+    // Show / meet-and-greet live data. The entertainments fetch is gated by its
+    // own catch above: this method is not wrapped in a try/catch by the base
+    // class, so an unguarded rejection here would reject the whole poll and take
+    // the ride live data down with it. A failing entertainments feed therefore
+    // drops only the shows. Today's performances are attached as showtimes and
+    // the show is CLOSED once the last one has passed.
     const nowMs = Date.now();
     const showLiveData: LiveData[] = [];
     for (const item of entData?.items ?? []) {
@@ -653,8 +655,14 @@ class PlopsaBase extends Destination {
 
   /** Per-show operating schedules from the entertainments feed. */
   private async buildShowSchedules(): Promise<EntitySchedule[]> {
-    const entResp = await this.fetchEntertainments();
-    const entData = (await entResp.json()) as PlopsaEntertainmentResponse;
+    let entData: PlopsaEntertainmentResponse;
+    try {
+      const entResp = await this.fetchEntertainments();
+      entData = (await entResp.json()) as PlopsaEntertainmentResponse;
+    } catch {
+      return [];
+    }
+
     const showSchedules: EntitySchedule[] = [];
 
     for (const item of entData?.items ?? []) {
--- a/src/parks/plopsa/plopsa.ts
+++ b/src/parks/plopsa/plopsa.ts
@@ -19,8 +19,8 @@ import config from '../../config.js';
 import {http, HTTPObj} from '../../http.js';
 import {CacheLib} from '../../cache.js';
 import {destinationController} from '../../destinationRegistry.js';
-import type {Entity, LiveData, EntitySchedule} from '@themeparks/typelib';
-import {formatDate, addDays, formatInTimezone} from '../../datetime.js';
+import type {Entity, LiveData, EntitySchedule, LiveTimeSlot} from '@themeparks/typelib';
+import {formatDate, addDays, formatInTimezone, constructDateTime} from '../../datetime.js';
 import dePanneLocations from './locations/plopsaland-de-panne.json' with {type: 'json'};
 
 // ── API response types ──────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


src/parks/plopsa/plopsa.ts
```diff
@@ -435,17 +483,17 @@ class PlopsaBase extends Destination {
   protected async buildLiveData(): Promise<LiveData[]> {
     const today = formatTodayInTimezone(this.timezone);
 
-    const [waitResp, poiResp, hoursResp] = await Promise.all([
+    const [waitResp, poiResp, hoursResp, entResp] = await Promise.all([
       this.fetchWaitTimes(),
       this.fetchPOI(),
       this.fetchTodayHours(today).catch(() => null),
+      this.fetchEntertainments(),
```
cubehouse: This `fetchEntertainments()` call has no `.catch()` guard, unlike `fetchTodayHours` right above it. Since `Destination.getLiveData()` doesn't wrap `buildLiveData()` in a try/catch, an entertainments-feed failure now rejects this whole `Promise.all` and takes ride wait-time live data down with it — the opposite of the PR's stated goal that shows surface "independently of the wait-times feed." Before this PR, entertainments wasn't fetched in `buildLiveData` at all, so this endpoint couldn't previously affect ride data at all.

Suggest gating it the same way: `this.fetchEntertainments().catch(() => null)`, then skip building show live-data when the result is null (mirrors how `hoursResp` is already handled a few lines down).

BeyondVertical: Good catch — you're right, and I've fixed it in 1902a3cb.

I confirmed the premise: `getLiveData()` (destination.ts:964-966) calls `buildLiveData()` with no try/catch, so the unguarded rejection propagated straight out and took the ride wait-time data with it. Before this branch `fetchEntertainments()` wasn't called here at all, so this was a regression I introduced.

Fixed exactly as you suggested:

```ts
this.fetchEntertainments().catch(() => null),
...
const entData = entResp ? (await entResp.json()) as PlopsaEntertainmentResponse : null;
```

The show loop already iterates `entData?.items ?? []`, so a null result simply drops the shows for that poll. I also rewrote the comment above that block, which claimed an independence that didn't actually exist.

**While verifying, I found the mirror image of the same defect** and fixed it in the same commit: `buildShowSchedules()` also called `fetchEntertainments()` unguarded, and `getSchedules()` (destination.ts:1055-1057) has no try/catch either — so an entertainments outage rejected `buildSchedules()` and discarded the **park calendar** too, contradicting the feed independence this branch claims. It now uses the same try/catch that `buildParkSchedule()` already had.

Both paths have regression tests, and I verified each one **fails without its guard** (`Error: entertainments 500`) rather than passing vacuously:

- `a failing entertainments feed drops only the shows, never the ride live data`
- `still returns the park schedule when the entertainments fetch fails`

Plopsa suite is 20/20, and both parks still pass the harness 4/4 (De Panne 102 entities / 67 live / 10 schedules, Deutschland 81 / 65 / 17).

BeyondVertical:
Rebased onto the latest `main` — merged and resolved the conflict with #249 (*treat 0-wait as no signal, report DOWN not OPERATING*), which rewrote the same `buildLiveData` ride-mapping block and renamed `plopsaDecideOperating` → `plopsaDecideStatus`.

**Both sides are kept:**
- **#249's ride semantics** — `plopsaDecideStatus`, `rawWait` (only a strictly positive reading counts as a live signal), `DOWN` for park-open + `temporarily_closed` + no signal, `STANDBY.waitTime = rawWait`.
- **This branch** — show/meet-and-greet live data, the `waitTimes ?? {}` relaxation, and the entertainments-fetch guards from your review.

The test file's import was the only textual conflict; #249's rewritten `plopsaDecideStatus` matrix and this branch's four new describe blocks are both retained (20/20 passing).

Verified live that the two features coexist correctly:

| Park | Harness | Attractions (live) | Shows (live) |
|---|---|---|---|
| Plopsaland Deutschland | 4/4 | 47/47 — 46 OPERATING, **1 DOWN** | 17/17, 14 with showtimes |
| Plopsaland De Panne | 4/4 | 53/53 — 47 OPERATING, **6 DOWN** (incl. SuperSplash) | 13/13, 6 with showtimes |

---

**On the red CI:** the failure is `src/__tests__/traceContextPropagation.test.ts > should propagate trace context through injection handlers` — `Test timed out in 35000ms`. It looks like a timing-sensitive flake rather than anything from this branch:

- This branch touches only `src/parks/plopsa/plopsa.ts` and its test file — no tracing, no HTTP layer.
- The test passes locally on this branch (2/2), but the file needs ~46s here; that test drives real HTTP with retry backoff and sits right on the 35s per-test timeout.
- Test caching can't be a factor either: `vitest.setup.ts` pins `CACHE_DB_PATH=':memory:'`, so nothing this branch adds touches a shared cache file.
- Everything else is green: 1471/1472 passing, and the whole plopsa suite is 20/20.

I don't have rerun rights on this repo — could you kick the job again? Happy to look deeper if it reproduces.

PR #271, opened 2026-07-31
chore: add containerised dev environment

lgnap:
Run the harness, tests and build without installing Node 24 / npm 11 on the host. The image pins the versions from package.json "engines", so contributors get the same toolchain regardless of what their distro ships.

- Dockerfile: node:24-slim, deps installed in their own layer. `npm ci` uses --ignore-scripts because the root "prepare": "tsc" hook cannot run before the sources are copied, and is unnecessary: tsx executes TypeScript directly.
- docker-entrypoint.sh: creates an empty .env when absent (every npm script runs `tsx --env-file=.env`, which aborts on a missing file) and restores node_modules on the first run with a bind mount.
- docker-compose.yml: bind-mounts the tree with :z for SELinux, keeps node_modules in a named volume so it never lands in the host tree.
- Makefile: self-documenting targets (make park PARK=efteling, make test, make web...). Defaults to podman-compose; override with `make COMPOSE="docker compose" dev` on a Docker host.
- .dockerignore: excludes caches, secrets and build output from the context.

Verified in-container: `make compile` clean, `make test` 1536 tests passing.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,17 @@
+.git
+node_modules
+web-ui/node_modules
+web-ui/dist
+dist
+coverage
+docs
+datacache
+snapshots
+.env
+cache.sqlite
+cache.sqlite-wal
+cache.sqlite-shm
+*.har
+http-dump.jsonl
+output.log
+testout_*
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+# Dev image for ParksAPI — Node 24 / npm 11 (see package.json "engines")
+FROM node:24-slim
+
+ENV NPM_CONFIG_UPDATE_NOTIFIER=false \
+    NPM_CONFIG_FUND=false \
+    TZ=UTC
+
+WORKDIR /app
+
+# Install deps first so the layer is cached independently of the source.
+# --ignore-scripts skips the root "prepare": "tsc" hook, which cannot run yet
+# (no sources at this point) and is not needed: tsx executes TypeScript directly.
+COPY package.json package-lock.json ./
+RUN npm ci --ignore-scripts
+
+# Baked-in sources, so the image is usable without a bind mount too.
+COPY . .
+
+COPY docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]
+CMD ["npm", "run", "dev"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,18 +3,35 @@ FROM node:24-slim
 
 ENV NPM_CONFIG_UPDATE_NOTIFIER=false \
     NPM_CONFIG_FUND=false \
-    TZ=UTC
+    TZ=UTC \
+    HOME=/home/node
 
 WORKDIR /app
+
+# node:24-slim ships an unprivileged `node` user (uid/gid 1000) but does not
+# select it. As root, everything the container writes into the bind mount is
+# root-owned on a Docker host — .env, dist/, cache.sqlite, docs/ — and the
+# developer then needs sudo to edit or delete their own files.
+#
+# Switching before `npm ci` rather than chowning afterwards keeps the image
+# ~200 MB smaller: a `RUN chown -R /app` would duplicate the whole
+# node_modules tree in a new layer. It also initialises the named node_modules
+# volume as uid 1000, so the entrypoint's `npm ci` fallback still works.
+#
+# Rootless Podman maps container uid 1000 to a subuid rather than to the
+# developer, so the Podman recipe in the Makefile header passes
+# --userns=keep-id; the entrypoint says so too if the mount is unwritable.
+RUN chown node:node /app
+USER node
 
 # Install deps first so the layer is cached independently of the source.
 # --ignore-scripts skips the root "prepare": "tsc" hook, which cannot run yet
 # (no sources at this point) and is not needed: tsx executes TypeScript directly.
-COPY package.json package-lock.json ./
+COPY --chown=node:node package.json package-lock.json ./
 RUN npm ci --ignore-scripts
 
 # Baked-in sources, so the image is usable without a bind mount too.
-COPY . .
+COPY --chown=node:node . .
 
 COPY docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
 ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,15 @@
 #   make build      → build the image once
 #   make park PARK=efteling
 #
-# Fedora ships Podman, not Docker. Override if you are on a Docker host:
-#   make COMPOSE="docker compose" dev
+# On a Podman host (Fedora and friends), override COMPOSE:
+#   make COMPOSE="podman-compose --podman-run-args=--userns=keep-id" dev
+#
+# The image runs as the unprivileged `node` user (uid 1000). Rootless Podman
+# maps that to a subuid rather than to you, so without --userns=keep-id the
+# container cannot write to the bind-mounted source tree. Docker needs no
+# such flag. Export COMPOSE in your shell if you use Podman day to day.
 
-COMPOSE ?= podman-compose
+COMPOSE ?= docker compose
 SERVICE  = parksapi
 RUN      = $(COMPOSE) run --rm $(SERVICE)
 
@@ -72,8 +77,8 @@ compile: ## Type-check / compile TypeScript to dist/
 	$(RUN) npm run build
 
 .PHONY: web
-web: ## Serve the web admin on http://localhost:8888
-	$(COMPOSE) run --rm --service-ports $(SERVICE) npm run web
+web: ## Serve the web admin on http://localhost:8888 (WEB_PORT to change)
+	$(COMPOSE) --profile web run --rm --service-ports web
 
 .PHONY: shell
 shell: ## Open a shell inside the container
@@ -90,8 +95,8 @@ clean: ## Remove containers and the node_modules volume
 	-$(COMPOSE) down -v
 
 .PHONY: clean-all
-clean-all: clean ## Also delete the built image
-	-podman rmi localhost/parksapi-dev
+clean-all: ## Also delete the built image (engine-agnostic, unlike a bare rmi)
+	-$(COMPOSE) down -v --rmi all
 
 ##@ Help
 
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,103 @@
+# ParksAPI — everything runs in a container (Node 24 / npm 11 inside).
+# Nothing here needs Node installed on the host.
+#
+#   make            → this help
+#   make build      → build the image once
+#   make park PARK=efteling
+#
+# Fedora ships Podman, not Docker. Override if you are on a Docker host:
+#   make COMPOSE="docker compose" dev
+
+COMPOSE ?= podman-compose
+SERVICE  = parksapi
+RUN      = $(COMPOSE) run --rm $(SERVICE)
+
+# Optional: extra flags forwarded to the test harness, e.g.
+#   make dev ARGS="--skip-schedules --verbose"
+ARGS ?=
+PARK ?=
+
+.DEFAULT_GOAL := help
+
+##@ Setup
+
+.PHONY: build
+build: ## Build the dev image (npm install happens here)
+	$(COMPOSE) build
+
+.PHONY: rebuild
+rebuild: ## Rebuild the image from scratch, ignoring layer cache
+	$(COMPOSE) build --no-cache
+
+.PHONY: deps
+deps: ## Reinstall node_modules inside the container volume
+	$(RUN) npm ci --ignore-scripts
+
+##@ Running the harness
+
+.PHONY: dev
+dev: ## Test every park (long: 76 destinations, hits live APIs)
+	$(RUN) npm run dev -- $(ARGS)
+
+.PHONY: park
+park: ## Test one park — make park PARK=efteling
+	@test -n "$(PARK)" || { echo "usage: make park PARK=<parkId>   (see: make list)"; exit 1; }
+	$(RUN) npm run dev -- $(PARK) $(ARGS)
+
+.PHONY: category
+category: ## Test a whole category — make category CATEGORY=Universal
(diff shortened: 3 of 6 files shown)


cubehouse:
Nice, thanks for this. I built and ran it: image builds in ~21s and `docker compose run --rm parksapi npm test` gives 1536/1536 passing. The `--ignore-scripts` reasoning on `npm ci` is right, and keeping `node_modules` in a named volume is the correct call.

Three things to fix before merge.

**1. The container runs as root, so it writes root-owned files into the host tree.**

`node:24-slim` does not set `USER`, so everything the container creates in the bind mount lands as `root:root` on the host. After a single run on my machine:

```
$ ls -la .env
-rw-r--r-- 1 root root 0 Aug  4 19:02 .env
```

The developer cannot then edit or delete their own `.env` without sudo, and the same applies to anything else the container writes into the mount: `dist/`, `cache.sqlite`, `docs/`.

Fix in the Dockerfile, after the `COPY . .`:

```dockerfile
RUN chown -R node:node /app
USER node
```

That also gets the named `node_modules` volume initialised as uid 1000 rather than root, so the entrypoint's `npm ci` fallback still works.

**2. `COMPOSE ?= podman-compose` is the wrong default for this repo.**

`make build` fails out of the box on a Docker host, which is most contributors. Please flip the default to `docker compose` and document the podman override instead:

```
make COMPOSE="podman-compose" dev
```

`clean-all` also hardcodes `podman rmi`, so it breaks the same way. `$(COMPOSE) down --rmi local` or a `$(CONTAINER_CLI) rmi` variable would cover both.

**3. `ports` at service level is not as inert as the comment suggests.**

The comment says the port is only published by `make web`, which holds for `compose run`, but a plain `compose up` publishes 8888 too. Since `make web` already passes `--service-ports`, the block can be dropped entirely, or moved behind a profile.

**Optional:** a short section in the README pointing at `make` would help, nothing currently mentions the container flow.

Once the ownership fix is in I will approve the workflow run so CI covers this branch.

cubehouse (changes requested):
Thanks for this, and sorry for the slow turnaround. The craft here is high: the comments explaining the uid mapping, the `:z` relabel and the `--ignore-scripts` rationale are better than most of what lands in this repo, and the core of it works. I built and ran it on Docker 29.6 and can confirm:

- `make build` succeeds, `make test` runs 1536/1536 passing in-container in 28s, `make list` and the help target work.
- The `USER node` change does what it claims: `.env`, `dist/` and `cache.sqlite` all come back owned by uid 1000, not root.
- `npm ci --ignore-scripts` is safe here. No native dependencies, and the cache uses the built-in `node:sqlite`.
- Dependency layer caching behaves correctly: touching a source file rebuilds only `COPY . .`, not `npm ci`.

I went through it fairly thoroughly, so there is a lot below. Three items stop it working at all, and the rest are worth fixing while you are in there. Happy to take this in stages if you would rather split it.

## Blockers

### 1. Compose refuses to start when a real `.env` is present

Compose automatically reads `./.env` for variable interpolation, and it requires strict `KEY=value` lines. A populated `.env` for this project contains multi-line values, which compose rejects:

```
$ docker compose config -q
failed to read <repo>/.env: line 12: unexpected character ";" in variable name <redacted>
```

Every `make` target then fails before a container is created. A `$` anywhere in a value also produces `The "xyz" variable is not set. Defaulting to a blank string.` This does not show up with the empty `.env` the entrypoint creates, which is why it passed your testing.

Fix: `COMPOSE ?= docker compose --env-file /dev/null` at `Makefile:16`, so the app's `.env` is used only by `tsx --env-file=.env` inside the container.

Related: if `TZ` or `WEB_PORT` ever appear in `.env` as application config, they silently change the container timezone and the published port.

### 2. The container plants a root-owned `node_modules/` in the host checkout

The named volume mounts at a path inside the bind mount, so the engine creates the mountpoint in the host tree as root before `USER node` applies. This is the exact failure `Dockerfile:11-19` says the PR exists to prevent, and it survives `make clean`:

```
$ rm -rf node_modules
$ docker compose run --rm --no-deps parksapi true
$ ls -ld node_modules
drwxr-xr-x 2 root root 4096 node_modules

$ npm ci        # back on the host
npm error code EACCES
npm error path .../node_modules/@babel
```

So anyone who tries the container once can no longer run npm on the host without sudo. Options: mount the volume somewhere outside the bind mount and point `NODE_PATH`/symlink at it, pre-create `node_modules` in the image so the mountpoint inherits `node:node`, or drop the named volume and accept in-tree modules.

### 3. `make web` cannot serve the UI

```
$ make web
$ curl -o /dev/null -w '%{http_code}' http://localhost:8888/            # 404
$ curl -o /dev/null -w '%{http_code}' http://localhost:8888/api/destinations   # 200
```

Root cause is partly ours: `npm run web` runs `npm run build` (which is `tsc`), never `web:build`, and `src/webServer.ts:457` serves `../web-ui/dist`. So a host clone 404s too. What the PR changes is that in the container it can never work: `.dockerignore:3-4` excludes `web-ui/node_modules` and `web-ui/dist`, and the named volume only covers `/app/node_modules`, so `cd web-ui && npm run build` fails with vite missing. It only serves a UI if the host already built one with host Node, which is the thing this section says you do not need.

Either wire a real web-ui install and build (needs a second named volume so `web-ui/node_modules`, about 200 MB, stays out of the host tree), or change `Makefile:80` and `README.md:93` to say API only.

## Should fix

### 4. Stale `node_modules` is guaranteed, not just possible

Because `npm ci` runs at build time, the image's `/app/node_modules` is non-empty, so Docker seeds the named volume from it on the first run. The guard at `docker-entrypoint.sh:22` (`[ -d /app/node_modules/tsx ]`) therefore always passes and the fallback branch is unreachable. A non-empty named volume is never re-seeded, so after a lockfile change plus `make build` the container keeps the old modules:

```
# after adding a dependency to package.json + package-lock.json and rebuilding
$ docker run --rm <new image> node -e "require('is-odd')"            # ok
$ docker compose run --rm parksapi node -e "require('is-odd')"       # Cannot find module 'is-odd'
```

No warning, no message. `make deps` fixes it and is not mentioned in the README. Better: write a hash of `package-lock.json` into the volume and re-run `npm ci` when it differs.

### 5. `make clean` exits 0 while doing nothing

`down -v` does not remove one-off `compose run` containers, and without `--profile web` it does not remove the `web` service either. Any survivor blocks the volume and the network, and the `-` prefix hides the failure:

```
$ docker compose --profile web up -d
$ docker compose down -v          # what make clean runs
 Volume  parksapi_node_modules  Resource is still in use
 Network default                 Resource is still in use
EXIT=0
$ docker ps -a --filter name=web  # still up, still holding 8888
```

Since `make clean` is the documented cure for finding 4, the cure is broken too. Use `down -v --remove-orphans` with `COMPOSE_PROFILES=web`, and drop the `-` so a failure is visible.

### 6. `make category` is broken for most categories

`Makefile:55` expands `$(CATEGORY)` unquoted, and `src/test.ts:50` reads only `args[categoryIdx + 1]`:

```
$ make -n category CATEGORY="Alton Towers"
docker compose run --rm parksapi npm run dev -- --category Alton Towers

$ npm run dev -- --category Alton Towers
No parks found in category: Alton
```

29 of the 64 registered categories contain a space. Fix is `--category "$(CATEGORY)"`. `$(PARK)` at `Makefile:50` has the same shape but no park id currently contains a space, so that one is cosmetic.

### 7. `.dockerignore` patterns are root-anchored, so nested files still get baked in

Docker matches patterns per path component, so `*.har` and `testout_*` only catch the context root. Verified by building a minimal context image and listing what arrives:

```
/ctx/src/parks/foo/capture.har     # not excluded
/ctx/src/testout_bar.json          # not excluded
/ctx/.vscode/settings.json
/ctx/.claude/settings.local.json
```

HAR captures can contain live request headers, so this one matters more than the rest. `.env`, root `*.har`, `cache.sqlite*`, `snapshots/`, `datacache/`, `node_modules` and `.git` are correctly excluded.

Fix: use `**/`-prefixed patterns and sync against `.gitignore`, which also has `scripts/`, `.vscode`, `.idea`, `.cursor`, `temp*.json`, `temp*.js`, `entity-map-viewer.html` and `opencode.json` that `.dockerignore` is missing.

### 8. A CRLF checkout bricks the container

`.gitattributes` currently pins only `*.js`, `*.cjs` and `*.mjs` to LF. A Windows contributor with `core.autocrlf=true` gets CRLF in `docker-entrypoint.sh`, and that is the same population this PR is aimed at:

```
$ docker run --rm <image with CRLF entrypoint>
exec /usr/local/bin/docker-entrypoint.sh: no such file or directory
EXIT=255
```

Three lines in `.gitattributes`: `*.sh text eol=lf`, `Dockerfile text eol=lf`, `Makefile text eol=lf`.

### 9. uid 1000 is hardcoded, and the error message blames Podman on Docker hosts

Anyone whose uid is not 1000 hits the failure path on plain Docker and gets podman-compose advice that does not apply:

```
$ docker compose run --rm --user 1001:1001 parksapi true
✗ cannot write to /app as uid 1001: the source tree is mounted from
  Podman: make COMPOSE="podman-compose --podman-run-args=--userns=keep-id" …
```

Worth a Docker branch in the message (`--user $(id -u):$(id -g)`, or a `UID` build arg). Also, I could not test Podman, but as written `--userns=keep-id` maps the host uid to the same uid in the container, so the documented recipe likewise only works when the developer's uid is exactly 1000. The general form is `--userns=keep-id:uid=1000,gid=1000` on podman 4.3+.

### 10. The writability guard only runs on a virgin checkout

`docker-entrypoint.sh:13` short-circuits once `.env` exists, so from the second run onward an unwritable tree fails as `SQLITE_CANTOPEN: unable to open database file`, which is exactly the "three commands later" failure the comment above it promises to prevent. Probe writability unconditionally: `touch /app/.env` when missing, otherwise write-probe a temp file in `/app`.

### 11. Nothing in CI exercises any of this

Six new root files that no workflow builds or runs will drift quietly: the floating base tag moves, a lockfile bump breaks `npm ci`, the `tsx`-presence probe breaks the day tsx stops being a direct dependency. A small `dev_container.yml` gated on `paths:` for these files (build, run the suite in-container, assert the entrypoint creates `.env`, assert no local-only files in the image, LF check) would keep it honest. Happy to hand you a starting point if useful.

## Minor and nits

- `Dockerfile:36`: `COPY` without `--chmod=0755`, so the entrypoint's mode comes from the build context. A checkout with `core.fileMode=false`, or a zip download, gives 644 and the container dies at exit 126 before any of your error messages can fire. `COPY --chmod=0755 …` fixes it (a `RUN chmod` will not, since `USER node` cannot chmod a root-owned file under `/usr/local/bin`).
- `Dockerfile:36`: this path already exists in `node:24-slim`, which ships its own `/usr/local/bin/docker-entrypoint.sh`. It works because line 37 sets an absolute `ENTRYPOINT`, but renaming to something like `parksapi-entrypoint.sh` avoids the surprise.
- `docker-compose.yml:6`: `image: parksapi-dev` is a fixed tag while the compose project name is derived from the directory. This repo gets worked in git worktrees, so `make build` in one tree silently replaces the image another is running, and `make clean-all` deletes it out from under them. Dropping the `image:` key lets compose namespace it per project.
- `Dockerfile:2`: `node:24-slim` is a floating tag, so two contributors building months apart get different 24.x. Fine for a dev image, but the README says the image "pins the versions", which it does not. Either pin `node:24.19-slim` or soften the wording.
- `Dockerfile:31`: about 53 MB of npm cache is left in the image, and the runtime path installs into a volume anyway. `npm ci --ignore-scripts && npm cache clean --force` in the same `RUN`.
- No `git` in the image, so `make shell` cannot commit or diff. Worth a README line, or an apt install.
- `README.md:108` says 75 destinations; `--list` reports 76, which matches the Makefile. Since the PR adds a section 20 lines above it, worth correcting in passing.
- `README.md:87` says every command above has a `make` equivalent, but `make build` builds the image while `npm run build` compiles TypeScript (that is `make compile`). It is the first line a reader tries.
- `Makefile:63`: `make health` drops `$(ARGS)`, so the containerised health check can never be scoped to a park or category even though `src/harness/health.ts` supports both.
- `WEB_ADMIN.md:32,171` say port 8080 while the app listens on 8888, and it makes `npm run web:dev` (Vite on port 3000) the recommended workflow, which has no container path at all: no target, no published port, no web-ui deps. That file was already stale, but the PR creates a direct contradiction between two docs.
- `docker-compose.yml:21`: the default command is `npm run dev`, so a plain `docker compose up` from a fresh clone fires roughly eight minutes of requests at 76 live park APIs from the contributor's IP. A `--list` default would be kinder.
- `docker-compose.yml:15`: `tty: true` with no `stdin_open`. Fine for `compose run`, but a stdin read under `compose up` blocks forever, and the tty puts ANSI escapes into piped logs.
- `docker-compose.yml:30`: `WEB_PORT` remaps only the host side, so with `WEB_PORT=9999` the app still prints `http://localhost:8888`.
- The README does not repeat that most parks need credentials in `.env`, and `make dev`'s help line does not mention that most will fail on the empty one the entrypoint creates. `make deps`, the manual fix for finding 4, is not in the README either. (`make park PARK=efteling` is a well-chosen first example, that one genuinely does pass on an empty `.env`.)
- The PR description says it defaults to podman-compose, but `Makefile:16` is `COMPOSE ?= docker compose`. The README is right; only the description is stale.

## Not a problem, so you do not have to chase it

The awk help block. `.*?##` is not portable notation, but the regex is only used as a line match and never for extraction, so greedy versus lazy cannot change the output, and every implementation treats the stray `?` as a no-op. Byte-identical output and rc=0 on gawk 5.2.1, mawk, one-true-awk and busybox awk. `%-14s` is also fine, the longest target name is 11 characters.

---

The blockers are 1, 2 and 3. If you want to split this, I would take a first pass with those plus the quoting fix in 6 and the `.gitattributes` lines in 8, and leave the rest for a follow-up. Thanks again for putting the work in.

lgnap:
too much work for a relative improvement, I close the PR ;-)

BeyondVertical:
I thought this was worth having, so I carried it on in #302. First commit is your state unchanged and co-authored to you; the rest works through the review point by point, so the diff against this PR stays readable.

Two notes that may interest you: the "pre-create `node_modules` in the image" option does not work — the engine creates the host mountpoint as root before image content or `USER` matter, verified — so the named volume is gone and the container runs as the host uid/gid instead. And `npm run web` now builds the UI, which is what made `/` 404 on a host clone too.

PR #272, opened 2026-08-03
fix(plopsa): stop serving a 12h-stale temporarily_closed flag

lgnap:
The POI feed is mostly cold metadata, but it is also the only carrier of schedule_info.temporarily_closed — live state that flips several times a day. Plopsaland's two splash rides are flagged closed until they open a few hours after the park does, and rides break down mid-afternoon.

At a 12h TTL a snapshot taken while a ride was open kept being served after it closed. plopsaDecideStatus reads !tempClosed as "open", so a stale FALSE surfaced as OPERATING with a 0-minute wait — the "Open, 0 min" artefact #249 set out to remove, reintroduced by the cache. The opposite direction, a stale TRUE after a ride opens, was already rescued by the hasWait priority.

Drop the TTL to 5 minutes: close enough to track the flag, still coarse enough to collapse the per-poll fetches of a collector running on a shorter interval. The feed is ~220 KB per language and `languages` holds one or two entries per park.

Verified live on 2026-08-03: DinoSplash (1327) and SuperSplash (90) both carried temporarily_closed with no wait while 34 of 54 rides had one, and the library emitted DOWN for exactly the six rides the feed flagged.

### DIFF
--- a/src/parks/plopsa/plopsa.ts
+++ b/src/parks/plopsa/plopsa.ts
@@ -278,7 +278,24 @@ class PlopsaBase extends Destination {
 
   // ── HTTP methods ──────────────────────────────────────────────
 
-  @http({cacheSeconds: 60 * 60 * 12})
+  /**
+   * The POI feed is mostly cold metadata (titles, images, height specs), but it
+   * is also the *only* carrier of `schedule_info.temporarily_closed` — live
+   * state that flips several times a day. Plopsaland's two splash rides open a
+   * few hours after the park does and are flagged closed until they do; rides
+   * also break down mid-afternoon.
+   *
+   * So this cannot be cached like metadata. It used to sit at 12h, which let a
+   * snapshot taken while a ride was open report OPERATING for the rest of the
+   * day after it closed — `plopsaDecideStatus` reads `!tempClosed` as "open",
+   * so a stale FALSE surfaces as "Open, 0 min". (The reverse, a stale TRUE
+   * after a ride opens, is already rescued by the `hasWait` priority.)
+   *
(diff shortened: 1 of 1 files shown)


cubehouse:
Thanks, the reasoning here checks out against the code. `schedule_info.temporarily_closed` really is only carried by the POI feed, and `plopsaDecideStatus` treats `!tempClosed` as OPERATING, so a stale FALSE does resurface as "Open, 0 min". A 12h TTL on a field that flips several times a day was the wrong shape.

Verified locally on this branch: `plopsaland` harness run passes 4/4, plopsa unit tests 23/23.

One change before this goes in.

`fetchPOI` has no `retries`, so it defaults to 0, and `buildLiveData` puts `fetchPOIAllLanguages()` in a bare `Promise.all` with no `.catch()`, unlike `fetchTodayHours` and `fetchEntertainments` which are both guarded. Since `buildLiveData` is not wrapped by the base class, a rejection there takes down the whole poll, not just the closed flags.

Going from ~2 POI fetches a day to ~288 multiplies the exposure to a transient upstream failure by roughly 144x, and each one now costs the park's entire live data for that poll. That is the same park-wide flapping failure mode the comments on `fetchTodayHours` already warn about.

Please add retries to match:

```ts
@http({cacheSeconds: 60 * 5, retries: 3})
async fetchPOI(language: string = this.apiLanguage): Promise<HTTPObj> {
```

Worth a line in the doc comment explaining that the retries exist because the shorter TTL puts this fetch on the live path, same as the note on `fetchTodayHours`.

With that, happy to merge.

lgnap:
done ;-)

PR #287, opened 2026-08-09
feat(energylandia): add Energylandia destination

cubehouse:
Adds Energylandia (Zator, Poland) — 89 attractions, live wait times, and 242 days of operating hours.

## Three backends, one park

| Concern | Source |
|---|---|
| Entities | Firestore `attractions`, `active == true`, `type == 'attraction'` |
| Schedules | Firestore `calendar/periods` |
| Wait times | A standalone host, not part of Firebase, joined by `queueTimeId` |

Auth is anonymous Firebase sign-in — the same thing the app does, so there's no account to maintain and nothing secret beyond the public Firebase web key. Every endpoint is config-only; the wait-time host is deliberately kept out of this repo, and with it unset the park still emits entities and schedules.

Closely follows the existing `flamingoland.ts` Firebase pattern (token caching, scoped `Bearer` injection, 401-invalidate, paginated collection reads).

## Two findings that shaped it

**`queueTimeId` is stored in two Firestore shapes.** 39 attractions hold it as `integerValue`, 50 as `stringValue` — 47 of those the empty string. Reading only `stringValue` gave a park that looked completely healthy:

```
✓ Found 89 entities      ✓ Found 89 live data entries
                           - With wait times: 3        ← should be 42
```

4/4 harness tests green, every ride OPERATING, and silently 3 wait times instead of 42. `fsId()` reads both shapes and treats the empty string as absent rather than coercing it — `Number('')` is 0, which would collide with the real counter id 0.

**The per-attraction `open` flag is not a live signal.** It tracks `active` almost exactly — every active attraction is flagged open, every inactive one closed — so it never moves across a day:

```
type=attraction  active=True   open=True   -> 84
type=attraction  active=False  open=False  -> 57
```

Trusting it would have published all 89 rides as OPERATING at 3am and right through the closed season. Live status is gated on the park's own published operating hours instead, with an unpublished date treated as open so a calendar gap degrades gracefully rather than blacking out a running park.

## Other decisions

- **Wait bounds are a garbage guard, not a plausibility filter** (`>= 0, < 600`, matching `gentingskyworlds`). The feed publishes odd-but-real values — one ride reports `310`, which the official app renders verbatim as "310 min". Suppressing that would misreport the park and would equally hide a genuine multi-hour queue.
- **No blocklist needed for operator rows.** Roughly a third of the feed is queue counters, FAST-PASS lanes and spares; they have no attraction to join to and drop out on their own.
- **Pagination is followed, not assumed.** `attractions` exceeds the 300-document page size, and a single-page read returns a truncated catalogue that looks perfectly valid.
- **Coordinates are deliberately absent.** They live in Proximiio via `proximiioId`; Firestore has no lat/lng, and taking on a second vendor for coordinates alone isn't worth the coupling here. The harness flags this as a known gap.

## Verification

Against both live backends, rebuilding expectations independently from the raw sources:

```
89 active attractions verified against the raw feed, 0 mismatches
orphan live ids: 0
schedule: 242 rows; today 10:00:00+02:00 -> 20:00:00+02:00
live: 89 entries, 42 with waits
```

33 gate tests, 746ms. Mutation-tested: reverting `fsId` to the string-only read fails 2 of them. Full suite green — 74 files, 1602 tests.

## Deploy note

Needs `ENERGYLANDIA_APIKEY`, `ENERGYLANDIA_PROJECTID` and the wait-time URL in the collector env; the ops template change is prepared separately.

### DIFF
--- a/src/parks/energylandia/__tests__/energylandia.test.ts
+++ b/src/parks/energylandia/__tests__/energylandia.test.ts
@@ -378,3 +378,51 @@ describe('Energylandia — attraction locations', () => {
     expect(ents.every((x: any) => x.location === undefined)).toBe(true);
   });
 });
+
+describe('Energylandia — pinned fallback coordinates', () => {
+  // Three rides are mapped in Proximiio but Firestore points at recreated
+  // features that no longer exist. The fallback is keyed by Firestore doc id
+  // and consulted ONLY on a miss, so a repointed CMS takes over again by
+  // itself.
+  const MINI_TRACK = 'JqDuKAgRzSP54oRShbuv';
+
+  function parkWith(features: any[]) {
+    const p: any = new Energylandia();
+    p.proximiioBaseUrl = 'https://geo.example';
+    p.proximiioToken = 'token';
+    p.getCalendarPeriods = async () => [];
+    p.getAttractionDocs = async () => [
+      doc(MINI_TRACK, {active: bool(true), type: str('attraction'), open: bool(true),
+        name: nameMap({EN: '220. Mini Track Tour Ride'}), proximiioId: str('org:stale')}),
+    ];
+    p.fetchProximiioFeatures = async () => ({json: async () => ({features})});
+    return p;
+  }
+
+  beforeEach(() => CacheLib.clear());
+  afterEach(() => CacheLib.clear());
+
+  test('fills in a ride whose proximiioId resolves to nothing', async () => {
+    const e: any = (await parkWith([]).getEntities()).find((x: any) => x.entityType === 'ATTRACTION');
+    expect(e.location).toEqual({latitude: 49.999318, longitude: 19.402042});
+  });
+
+  test('live Proximiio data wins over the pin, so a CMS fix self-heals', async () => {
+    const live = [{id: 'org:stale', geometry: {type: 'Point', coordinates: [19.5, 50.5]}}];
+    const e: any = (await parkWith(live).getEntities()).find((x: any) => x.entityType === 'ATTRACTION');
+    expect(e.location).toEqual({latitude: 50.5, longitude: 19.5});
+  });
+
+  test('every pinned coordinate sits inside the park', async () => {
+    // Guards against a fat-fingered digit or a transposed pair being pasted in.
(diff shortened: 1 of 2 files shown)


cubehouse:
## Coordinates added — 0 → 86 of 89 attractions

Source is **Proximiio**, the mapping vendor behind the app's own park map. Firestore carries no lat/lng at all, only a `proximiioId` per attraction — and it's stored in exactly the `"<organizationId>:<featureId>"` form Proximiio's API returns, so the join is direct with no massaging.

Auth is a static application JWT with no `exp` claim (it identifies the park's Proximiio application, not a user), so there's no refresh flow. Base URL and token are both config; unset, entities emit without coordinates rather than not at all.

### Two traps in that collection

**Most of it isn't attractions.** Of 1036 features, 625 are LineStrings — the wayfinding path network — and 3 are Polygons. Only the 408 Points carry a usable position. A LineString's `coordinates` is an *array of pairs*, so reading `[0]`/`[1]` off one yields arrays where numbers are expected and produces a garbage location rather than an error.

**GeoJSON is longitude-first**, the reverse of how it's consumed. Transposed, Energylandia lands off the coast of Somalia. The mapping is explicit rather than positional, and a test pins the real orientation.

Null-island `(0,0)` and out-of-range pairs are rejected — an exactly-zero pair is far more likely to be missing data than a real position.

### Verification

```
attractions: 89, with location: 86, missing: 3
outside the park bounding box: 0        ← proves no transposition
Formuła Autodrom    50.000026, 19.404603
Tiki Ride           49.997757, 19.409181
```

The 3 misses (Mini Track' Tour Ride, Candy Critters, Candy Carousel) have `proximiioId`s that aren't in Proximiio's feature set — newer rides not yet mapped. They keep their wait times and simply have no location, rather than being dropped.

Harness now reports `⚠ Missing location: ATTRACTION:3` instead of `:89`.

43 gate tests (822ms), full suite 1612 passing.

### Deploy note

Adds two more config values, `ENERGYLANDIA_PROXIMIIOBASEURL` and `ENERGYLANDIA_PROXIMIIOTOKEN`, alongside the three already needed.

PR #289, opened 2026-08-09
feat(energylandia): publish showtimes, shows and restaurants

cubehouse:
Branched from `main`, independent of #288.

## What

The park keeps performances in a `shows` collection of its own, separate from the `attractions` documents the implementation already read. 13 published shows, 259 performances across a full week, each with a time, a venue name and usually a link to the venue's document.

Also emits the 72 restaurants that were being filtered out of `attractions`. Shops and games stay unpublished — there is no entity type for them.

| | Before | After |
|---|--:|--:|
| ATTRACTION | 89 | 89 |
| RESTAURANT | 0 | 72 |
| SHOW | 0 | 13 |
| **Total entities** | **91** | **176** |

## The thing this had to get right

**The timetable is keyed by weekday, not by date.** It is a recurring pattern with no notion of the calendar, so read as-is it will happily describe Saturday's parade on a Saturday in the closed season. It is gated on the operating calendar exactly as wait times already are: outside published hours a show reports CLOSED and emits no times at all.

An empty calendar still degrades to open rather than blacking out a running park — the rule attractions already follow. One policy for that failure, not two. Both paths are tested.

## Four things the data required

Each of these was found by checking, and two of them were bugs caught by the tests rather than by reading.

**`timeEnd` is null on all 259 slots.** End times come from the show's own stated `duration` (10 to 120 minutes across the collection) — the park's number, not a guess. A show without a usable one gets a null end rather than an invented one.

**Both timestamps must render the same way.** Taking the end off the shifted `Date` and calling `.toISOString()` is correct to the instant but writes it as UTC, so a single showtime carried `…T12:45:00+02:00` next to `…T11:00:00.000Z`. It now round-trips through the same `constructDateTime` as the start, which also puts a past-midnight performance on the next date.

**Slot times are range-checked, not shape-checked.** `\d{2}:\d{2}` accepts `25:99`, which `constructDateTime` resolves by rolling into the following day — publishing a performance at an hour the park never stated, on a date it never scheduled.

**Slots are sorted by time.** The CMS stores entry order, which is not chronological: one show lists 16:30 before 13:30. Unsorted, "the day's last performance" is the wrong entry, which is what decides whether the show still counts as running. A show now goes CLOSED once its last performance has finished rather than staying OPERATING until the park shuts.

## Venues

63 of 259 slots omit `attractionId` entirely while still naming their venue in `place`, so a missing link is not a missing venue and those slots are not treated as disagreement. Twelve of thirteen shows sit at one venue all week and inherit its coordinates; "Meeting with Mascots" rotates between three and gets no location rather than whichever happened to be listed first.

Show entity ids are namespaced `energylandia-show-<docId>`: a Firestore document id is unique only *within* its collection, so an unprefixed id would let `shows/abc` overwrite `attractions/abc`. Attraction ids are unchanged.

## Verification

Harness 4/4 against the live backend. Suite goes 64 → 100 tests for this park, 1669 passing overall, `tsc` clean.

Sampled at 14:00 park-local on an operating Sunday: all 13 shows report, 37 performances between them, every one carrying both a start and an end.

```
Crew! All aboard!              OPERATING dur=15 venue=lzCCNSqL slots=3
      2026-08-09T12:45:00+02:00  ->  2026-08-09T13:00:00+02:00
      2026-08-09T14:15:00+02:00  ->  2026-08-09T14:30:00+02:00
      2026-08-09T16:15:00+02:00  ->  2026-08-09T16:30:00+02:00
PARADE                         OPERATING dur=15 venue=9eTCroaC slots=1
      2026-08-09T19:45:00+02:00  ->  2026-08-09T20:00:00+02:00
...
shows=13 OPERATING=13 slots=37 withEndTime=37
at 21:30 local (after last performance): shows still OPERATING = 0
```

The evening parade ending at exactly 20:00, the park's published closing time, is a good sign the durations are real.

## Note

Existing test mocks needed a `getShowDocs` stub. Without one, `getEntities()` attempted a real request and 20 tests hung to timeout — the new collection read is fatal to the emission if it fails, deliberately, matching how `attractions` already behaves. Publishing a park with its shows silently removed is a deletion event, and a thrown error is never cached so it self-heals on the next poll.

### DIFF
--- a/src/parks/energylandia/__tests__/energylandia.test.ts
+++ b/src/parks/energylandia/__tests__/energylandia.test.ts
@@ -9,6 +9,13 @@ import {
   isWithinOperatingWindow,
   parkLocalDateTime,
   buildLocationIndex,
+  parkLocalWeekday,
+  parseShowSlots,
+  parseDurationMinutes,
+  buildShowtimes,
+  showStatusFromShowtimes,
+  resolveShowVenueId,
+  WEEKDAYS,
 } from '../energylandia.js';
 import {CacheLib} from '../../../cache.js';
 
@@ -202,6 +209,7 @@ describe('Energylandia — live data', () => {
 
   function park(opts: {days?: string[]; open?: string; close?: string} = {}) {
     const p: any = new Energylandia({config: {...BLANK_CONFIG, waitTimesUrl: 'https://feed.example/'}});
+    p.getShowDocs = async () => [];
     p.getAttractionDocs = async () => ATTRACTIONS;
     p.fetchWaitTimes = async () => ({text: async () => JSON.stringify(FEED)});
     p.getCalendarPeriods = async () => [{
@@ -215,10 +223,22 @@ describe('Energylandia — live data', () => {
   beforeEach(() => CacheLib.clear());
   afterEach(() => CacheLib.clear());
 
-  test('only active attractions become entities — restaurants and retired rides are excluded', async () => {
+  test('only active rides become ATTRACTIONs — a restaurant is not one, a retired ride is not published', async () => {
     const entities = await park().getEntities();
     const names = entities.filter((e: any) => e.entityType === 'ATTRACTION').map((e: any) => e.name).sort();
     expect(names).toEqual(['No Counter', 'Pepsi Hyperion', 'Tsunami Drop']);
+  });
+
+  test('a restaurant becomes a RESTAURANT rather than being dropped', async () => {
+    const entities = await park().getEntities();
+    const dining = entities.filter((e: any) => e.entityType === 'RESTAURANT');
+    expect(dining.map((e: any) => e.name)).toEqual(['Pizzeria']);
+    expect(dining[0].parkId).toBe('energylandia-park');
+  });
+
+  test('an inactive document is published under no entity type at all', async () => {
+    const entities = await park().getEntities();
+    expect(entities.map((e: any) => e.id)).not.toContain('energylandia-x1');
   });
 
   test('joins the feed by queueTimeId across BOTH id shapes', async () => {
@@ -391,6 +411,7 @@ describe('Energylandia — attraction locations', () => {
     const p: any = new Energylandia({config: {
       ...BLANK_CONFIG, proximiioBaseUrl: 'https://geo.example', proximiioToken: 'token',
     }});
+    p.getShowDocs = async () => [];
     p.getAttractionDocs = async () => ATTRACTIONS;
     p.getCalendarPeriods = async () => [];
     p.fetchProximiioFeatures = async () => ({
@@ -444,6 +465,7 @@ describe('Energylandia — pinned fallback coordinates', () => {
       ...BLANK_CONFIG, proximiioBaseUrl: 'https://geo.example', proximiioToken: 'token',
     }});
     p.getCalendarPeriods = async () => [];
+    p.getShowDocs = async () => [];
     p.getAttractionDocs = async () => [
       doc(MINI_TRACK, {active: bool(true), type: str('attraction'), open: bool(true),
         name: nameMap({EN: '220. Mini Track Tour Ride'}), proximiioId: str('org:stale')}),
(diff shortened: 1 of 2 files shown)


cubehouse:
## Review round: four sub-agents, findings applied

Pushed `760df00b` + `b39b4fc8`. Tests 100 → 114, suite 1683 passing, harness still 4/4.

### Fixed

**A `shows` outage no longer takes the attractions' live data down with it.** The shows read sat in `buildLiveData`'s `Promise.all`, and `readCollection` throws on an empty collection by design — so the park renaming `shows`, tightening a rule on it, or emptying it out of season would have stopped all 89 attractions publishing status and wait times, for as long as it lasted. Confirmed by fault injection, not by reading. My original note defended this as deliberate; that reasoning holds for `getEntities()`, where a missing entity is a downstream deletion, but not for `getLiveData()`, where a missing show row merely goes stale. I had the harsher response on the milder failure. `getActiveShowsSafe` splits them, matching the trade `plopsa.ts:611` documents and the split this file already makes for Proximiio and the wait feed. Both directions are now tested.

**Showtimes are gated on the day, not the instant.** Gating the times themselves on the operating window left the park with no showtimes for ~14 hours out of every 24, all 13 shows appearing at once at 10:00. Times are reference information about the day; `status` still reflects the instant. Verified live at 21:28 Warsaw, 88 minutes after close: 13 shows carrying their day's times, all `CLOSED`.

**End times are formatted from the instant.** Round-tripping through a park-local wall clock and re-resolving it discards the offset, and on the autumn fold that clock is ambiguous — `constructDateTime` picks the post-transition occurrence, turning a 20-minute show starting 01:50 into an 80-minute one. Unreachable with current data (latest slot 20:15, longest duration 20 min), but wrong, and `formatInTimezone` is simpler than what it replaced. Both DST directions now have tests.

**`duration` and the slot `attractionId` now use `fsId`, not `fsString`.** This CMS stores identifiers as `integerValue` and `stringValue` interchangeably. Reading one shape is exactly how wait times once published for 3 rides instead of 42 — and on `duration` it would have silently nulled every end time. I used the right reader on one and the wrong one on the other, and pinned neither.

Also: the slot-level `active` flag is honoured (no slot sets it false today); nameless shows are skipped in live data as they already were in entities, which was shipping an orphan row for an entity nobody published; `type` becomes `'Performance Time'`, the form 8 of 10 parks here emit and which reaches the public API verbatim; the unread `place` field is dropped.

### Deliberately not fixed

**The timetable and the calendar genuinely disagree.** Over the next 60 operating days, 8.5% of slots start after the published close — a Tuesday parade at 20:15 against a 20:00 close, and 19:00 shows against the 18:00 close the autumn calendar uses. Neither document is self-evidently wrong: a closing parade plausibly runs after the stated close, while a 19:00 show on an 18:00 day looks like a timetable nobody updated. Censoring the park's own showtimes with its other document would drop real performances, so they publish as stated and the conflict is `console.warn`ed once per build rather than discovered downstream. Happy to switch to hard filtering if you'd rather.

### Two tests were weaker than they read

The duplicate-id check ran against a fixture whose only attractions were `type: 'show'`, so the emission held no rides and it compared shows against themselves. And nothing pinned the weekday to park time, because every fixture clock had the same weekday in UTC and Warsaw. Both now use fixtures that fail when the behaviour breaks.

Ten mutations tried against the suite, all ten caught. The last one to fall was my own new venue test — its fixture had no slot on the simulated day, so a single-day read also produced no location and the assertion passed for the wrong reason (`b39b4fc8`).

### Verified, unchanged

Every numeric claim in the code comments was checked against the live backend: 157 show docs / 13 active, 259 slots, `timeEnd` null on all 259, 63 missing `attractionId` each still carrying `place`, 12-of-13 single-venue, 316 attraction docs / 89 rides / 72 restaurants. Nothing to correct. Entity id stability, hierarchy, cache keys and the live-data contract all came back clean, as did a brute-force check of 131,400 instants around local midnight confirming the weekday and date cannot disagree.

PR #292, opened 2026-08-11
feat(dlp): surface attractions hidden from Disney mobile app

Kaitology:
## Summary
- Replace per-ride DLP visibility exceptions with a `hideFunctionality` policy map so app-hidden attractions surface in entity and live feeds.
- Attractions flagged `Hide from Mobile App` or `Hide from Web List + Mobile App` now pass POI filtering; service-hidden non-attractions stay filtered.
- `SERVICE_HIDE_EXCEPTIONS` keeps only shows that must bypass service hide (e.g. Live Your Story / P1GS93).

## Test plan
- [x] `npm test -- src/parks/dlp/__tests__/visibility.test.ts` (5 tests)
- [ ] After deploy: confirm app-hidden DLP rides (e.g. Mickey''s PhilharMagic, Frozen Ever After) appear in `/v1/entity/{park}/live`

### DIFF
--- a/src/parks/dlp/__tests__/visibility.test.ts
+++ b/src/parks/dlp/__tests__/visibility.test.ts
@@ -11,30 +11,6 @@ function stubbedPark(): DisneylandParis {
   const park = new DisneylandParis();
   vi.spyOn(park as any, 'getPOIData').mockResolvedValue({
     ThemePark: [{id: 'P1', name: 'Disneyland Park', type: 'ThemePark'}],
-    Attraction: [
-      {
-        id: 'P1DA10',
-        name: 'Disneyland Railroad Discoveryland Station',
-        type: 'Attraction',
-        location: {id: 'P1'},
-        hideFunctionality: 'Hide from Web List + Mobile App',
-      },
-      {
-        id: 'P1NA16',
-        name: 'Disneyland Railroad Fantasyland Station',
-        type: 'Attraction',
-        location: {id: 'P1'},
-        hideFunctionality: 'Hide from Web List + Mobile App',
-      },
-      {
-        // Control: retirement-flagged record not in VISIBILITY_EXCEPTIONS — must stay dropped.
-        id: 'P1DA14-OLD',
-        name: 'Retired Attraction',
-        type: 'Attraction',
-        location: {id: 'P1'},
-        hideFunctionality: 'Hide from Web List + Mobile App',
-      },
-    ],
     Entertainment: [
       {
         id: 'P1GS93',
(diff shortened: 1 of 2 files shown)


cubehouse (changes requested):
Thanks for this. Before merging I diffed the old and new filters against the live POI feed, and the results don't line up with the description.

The change has exactly one effect: attractions flagged `Hide from Web List + Mobile App` now pass. That adds 7 entities, 129 -> 136:

- `P1DA10` Disneyland Railroad Discoveryland Station (has wait data)
- `P1NA16` Disneyland Railroad Fantasyland Station (has wait data)
- `centralPlazaHub` Central Plaza
- `old-P2EA02`, `P1MA06-OLD`, `P1NA17-OLD`, `P1DA14-OLD` (retired records, no live data)

That flag is effectively Disney's retirement marker. Every OLD-ish ID in the feed carries it, 15 of them across Attraction, Entertainment and Restaurant, and we publish none of them today. Bypassing it category-wide is what lets the dead records in.

Three assumptions in the PR that don't hold against the feed:

1. Mickey's PhilharMagic (`P1DA13`) is flagged `Hide from the Service`, not app-hidden, so this PR still drops it. The test only passes because the fixture stubs it as `Hide from Mobile App`. It also already ships today as `P1G103` with `entityType: SHOW`.
2. Frozen Ever After (`P2EA00`) and Tangled Spin (`P2DA00`) have an empty flag right now, so removing them from `VISIBILITY_EXCEPTIONS` is a no-op that quietly drops their guard if Disney re-flags them.
3. `Hide from Mobile App` is not a string the API emits. The real value is `Hide from the Mobile App` (44 rows), which matches no policy entry and falls through `!policy`.

What I would merge: keep `HIDE_RULES` and `VISIBILITY_EXCEPTIONS` as they are, and just add `P1DA10` and `P1NA16` to the exceptions. Those two stations are the only live attractions wearing the flag, so that captures the entire real win in two lines with no fallout. Happy to take it from here if you'd rather.

cubehouse (approved):
Verified against the live feed. Entities 129 -> 131, live rows 67 -> 69, and the delta is exactly the two stations:

- `P1DA10` Disneyland Railroad Discoveryland Station, live `DOWN`
- `P1NA16` Disneyland Railroad Fantasyland Station, live `CLOSED`

Retirement-flagged records stay out (no `old-` / `-OLD` IDs published, `centralPlazaHub` still filtered), and `P2EA00` / `P2DA00` / `P1GS93` keep their existing guards. `npm test src/parks/dlp/__tests__/visibility.test.ts` passes, 3/3, and the `P1DA14-OLD` control is a good addition.

Thanks for turning this around so quickly.

cubehouse (changes requested):
The runtime change is still right, and I verified it against the live feed on the previous head: entities 129 -> 131, live rows 67 -> 69, delta is exactly `P1DA10` (DOWN) and `P1NA16` (CLOSED), with retirement-flagged records staying out. That part is unchanged here.

But `c143646c` dropped the test coverage that `bd2bcbb0` added. `visibility.test.ts` is now untouched relative to `main`, so nothing in the suite would catch a regression on this.

"Two lines" was about the runtime logic, not about shipping without tests. Please restore the test additions from `bd2bcbb0`:

- the `P1DA10` / `P1NA16` fixtures asserting both surface as `ATTRACTION`
- the `P1DA14-OLD` control asserting retirement-flagged records still drop

The control is the valuable half. It is the thing that stops someone widening this back into a category-wide bypass later.

Worth restoring the fuller comments from that commit too ("retirement-flagged but still reports wait times"). Without the rationale, the next reader sees two station IDs in an exception list with no reason to keep them.

Happy to push the test commit myself if that is easier.

cubehouse (approved):
Tests restored in dadd437d and the rationale is back in the comment. Full suite green locally (74 files, 1693 tests), and I confirmed the new test fails when the two IDs are removed, so it is a real regression guard. Thanks for the quick iterations on this one.

Kaitology:
Thanks for the follow-up! I was still sleeping but again you're much faster than me. Much appreciated cube.

PR #293, opened 2026-08-11
feat(dlp): publish all shows and give showtimes a real end time

BeyondVertical:
## Summary

Two changes to Disneyland Paris entertainment data:

- **Character meet & greets are published as shows.** They were dropped by `mapEntityType` because their `subType` was missing from `SHOW_SUBTYPES` — nothing to do with the hide rules.
- **Showtimes with a published POI `duration` get a real end time.** The schedule feed reports every `PERFORMANCE_TIME` slot with `endTime` equal to `startTime`, so a window has to come from the POI `duration`; shows without one keep their zero-length slot.

| | `main` | this branch |
|---|---|---|
| Entities | 127 | **147** |
| `SHOW` | 23 | **43** |
| Showtime slots | 90 | **177** |
| …with a real window | 0 | **41** (the rest publish no `duration`) |
| DLP tests | 2 | **202** |

`ATTRACTION` (51), `RESTAURANT` (50) and `PARK` (2) are unchanged. No new HTTP request — `duration` and the meet & greet records ride along on the POI query that already runs.

Both columns measured back to back against `main` at `1c8a739d`, which is merged in here. Slot totals move through the day as performances pass.

### What this costs, and what it does not fix

Two things worth having in front of you rather than in a thread:

- **A cold `getEntities()` now runs the 60-day schedule sweep**, because the meet & greet gate needs to know what has performances. Measured here, both back to back: **43s cold against ~1s on `main`**, warm 7ms. It is the same 60 days `buildSchedules` already fetches and `getScheduleForDate` caches for 12h, so a full pass pays it once instead of twice — but a consumer that only ever calls `getEntities()` pays a sweep it never asked for. Reported as high as 75s depending on the upstream.
- **The 20 new entities carry no coordinates.** Disney publishes no location for meet & greets, so there is nothing to fix here, but the harness warning goes `SHOW:20` to `SHOW:40`.

## Meet & greets

59 in-park records carry `subType: "Character Experience - Meet & Greet"`. 17 pass the existing hide rules unchanged.

Three more are added to `VISIBILITY_EXCEPTIONS`. All three are flagged `Hide from Web List + Mobile App`, but the schedule feed lists them on nearly every day of the next 60:

| ID | days with performances | name |
|---|---|---|
| `P2MG31` | 51 / 60 | Meet Goofy, the Movie Director |
| `P1MG21` | 46 / 60 | An Encounter with Captain Hook |
| `P1MG05` | 46 / 60 | Meet Donald Duck or his friends |

This is deliberately narrow, and it does not reopen the question settled in #292. I cross-referenced all 128 in-park `Entertainment` records against 60 days of the feed:

- Of the 33 hide-filtered show subtypes (Stage Show / Fireworks / Parade / Atmosphere), exactly **one** has an upcoming performance — `P1GS93`, already an exception. The other 32 are seasonal-out-of-window (10, e.g. Christmas Parade, Halloween Celebration) or retired/test records (22, including literal `FakeID` and `asgardPedestral`). The retirement-marker reading holds for shows, so the hide rules stay as they are.
- For meet & greets it does not hold, which is why these three are listed individually rather than by relaxing a rule.

No `-old` / `-OLD` / `old-` record enters the set. Magic Shots (13) stay out: they are PhotoPass photo spots where the character is composited into the picture afterwards — Disney's own copy says *"The Disney Character is not physically present"* — and they have no duration, no coordinates and no schedule entry on any of the 60 days checked.

Six of the 20 publish no performance times because they run on the virtual queue (`virtualQueue: true`): Hero Training Center, Meet Mickey Mouse, Princess Pavilion, Rencontre Royale, Meet a Toy Story Character. Starport is a walk-up meet with no published times. They are real, operating experiences and are emitted as entities without showtimes.

## Showtime durations

`duration` lives on the concrete `Entertainment` type, not the shared `Activity` interface, so it needs an inline fragment — requesting it from the common selection fails schema validation.

16 of 44 shows publish a duration; the remaining 28 keep a zero-length slot, which matches how meet & greets and drop-in formats such as Animation Academy actually work.

```
P1GS21  11:30 +33min   Disney Stars on Parade
P1GS99  22:00 +22min   Disney Tales of Magic
P1GS34  12:30 +30min   The Lion King: Rhythms of the Pride Lands
P2GS78  10:00 +28min   Disney Princess Cavalcade
```

One defect fixed and one hardening along the way:

- Durations were read from state only `buildEntityList` populated, so they were empty whenever `getLiveData()` ran on its own. They now come from the POI records `buildLiveData` already loads.
- Lengths are parsed rather than trusted. This is hardening of the new path, not a shipped-bug fix: on `main` the old `(minutes || 0) + ((hours || 0) * 60)` expression was unreachable because `fetchPOI` never requested `duration` — the operative change is adding the field to the query. Numeric strings are coerced (the old expression would have concatenated them), and negative or longer-than-a-day values fall back to the feed's end time.

`getPOIData` moves to `cacheVersion: 5` because the cached shape gains a field. #295 took 4 and the two query texts differ, so they need separate keys.

## Test plan

- [x] `npm test src/parks/dlp` — 202 tests across 10 files; full suite 1955 across 89 files
- [x] `npx tsc --noEmit`
- [x] `npm run dev -- disneylandparis --clear-cache` — 4/4, entities 147, zero orphans in both directions
- [x] New `showtimes.test.ts` verified to fail against the previous code (`expected +0 to be 30`), including a case that calls `getLiveData()` without a preceding `getEntities()`
- [x] Duration parsing fuzzed through the public path: missing, empty, null members, negative, non-numeric, numeric strings, longer than a day

### DIFF
--- a/src/parks/dlp/__tests__/scheduleSweepCache.test.ts
+++ b/src/parks/dlp/__tests__/scheduleSweepCache.test.ts
@@ -1,0 +1,69 @@
+import {describe, it, expect, vi, afterEach} from 'vitest';
+import {DisneylandParis} from '../disneylandparis.js';
+import {CacheLib} from '../../../cache.js';
+
+/**
+ * The 60-day sweep behind the meet & greet gate is reached from all three
+ * public entry points, so it has to survive in the cache — including when the
+ * schedule feed is down, which is when re-running it is most expensive.
+ *
+ * CacheLib reads a cached null back as a miss, so the outage cannot be
+ * signalled by returning null; it travels as `answered: false` alongside the
+ * ids instead. These tests count calls into getScheduleForDate, which is what
+ * a real outage would turn into failed upstream requests.
+ */
+describe('DLP schedule sweep caching', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+    CacheLib.clearAll();
+  });
+
+  const sweepingPark = (rows: unknown[]) => {
+    CacheLib.clearAll();
+    const park = new DisneylandParis();
+    const fetch = vi.spyOn(park as any, 'getScheduleForDate').mockResolvedValue(rows);
+    return {park, fetch};
+  };
+
+  it('sweeps once across repeated calls while the feed is down', async () => {
+    const {park, fetch} = sweepingPark([]);
+
+    const first = await (park as any).getScheduledActivityIds();
+    const perSweep = fetch.mock.calls.length;
+    expect(perSweep).toBeGreaterThan(0);
+
+    await (park as any).getScheduledActivityIds();
+    await (park as any).getScheduledActivityIds();
+
+    // The assertion that matters: three calls, one sweep. A null return here
+    // would read back as a cache miss and make it three.
+    expect(fetch.mock.calls.length).toBe(perSweep);
+    expect(first).toEqual({answered: false, ids: []});
+  });
+
+  it('sweeps once across repeated calls while the feed is healthy', async () => {
+    const {park, fetch} = sweepingPark([
+      {id: 'P1MG05', schedules: [{startTime: '10:00:00', endTime: '10:30:00'}]},
+      {id: 'P1MG99', schedules: []},
+    ]);
+
+    const first = await (park as any).getScheduledActivityIds();
+    const perSweep = fetch.mock.calls.length;
+
+    await (park as any).getScheduledActivityIds();
+
+    expect(fetch.mock.calls.length).toBe(perSweep);
+    expect(first.answered).toBe(true);
+    // A row with no performances is not a reason to publish the entity.
+    expect(first.ids).toEqual(['P1MG05']);
+    expect(fetch.mock.calls.length).toBe(perSweep);
+  });
+
+  it('reports the outage rather than an empty estate', async () => {
+    const {park} = sweepingPark([]);
+    const withData = await (park as any).getMeetAndGreetIdsWithData();
+    // null is the caller's "publish all of them"; an empty Set would unpublish
+    // every meet & greet for the 12h the sweep is cached.
+    expect(withData).toBeNull();
+  });
+});
--- a/src/parks/dlp/__tests__/showtimes.test.ts
+++ b/src/parks/dlp/__tests__/showtimes.test.ts
@@ -1,0 +1,158 @@
+import {describe, it, expect, vi, beforeEach} from 'vitest';
+import {DisneylandParis} from '../disneylandparis.js';
+
+/**
+ * The schedule feed reports every PERFORMANCE_TIME slot with endTime equal to
+ * startTime, so a show's window has to come from its POI duration. Shows
+ * without a published duration keep the feed's own endTime.
+ */
+function stubbedPark(): DisneylandParis {
+  const park = new DisneylandParis();
+
+  vi.spyOn(park as any, 'getPOIData').mockResolvedValue({
+    ThemePark: [{id: 'P1', name: 'Disneyland Park', type: 'ThemePark'}],
+    Entertainment: [
+      {
+        id: 'P1GS21',
+        name: 'Disney Stars on Parade',
+        type: 'Entertainment',
+        subType: 'Parade',
+        location: {id: 'P1'},
+        duration: {hours: 0, minutes: 30},
+      },
+      {
+        id: 'P1GS99',
+        name: 'Disney Tales of Magic',
+        type: 'Entertainment',
+        subType: 'Fireworks',
+        location: {id: 'P1'},
+        duration: {hours: 1, minutes: 5},
+      },
+      {
+        // No duration published — must fall back to the feed's endTime.
+        id: 'P2GS54',
+        name: 'Animation Academy',
+        type: 'Entertainment',
+        subType: 'Stage Show',
+        location: {id: 'P1'},
+      },
+    ],
+  });
+
+  vi.spyOn(park as any, 'getWaitTimes').mockResolvedValue([]);
+  vi.spyOn(park as any, 'getPremierAccess').mockResolvedValue([]);
+  vi.spyOn(park as any, 'getVirtualQueueData').mockResolvedValue([]);
+  vi.spyOn(park as any, 'getScheduleForDate').mockImplementation(async (date) => [
+    {
+      id: 'P1GS21',
+      schedules: [{date, startTime: '17:30:00', endTime: '17:30:00', status: 'PERFORMANCE_TIME'}],
+    },
+    {
+      id: 'P1GS99',
+      schedules: [{date, startTime: '22:45:00', endTime: '22:45:00', status: 'PERFORMANCE_TIME'}],
+    },
+    {
+      id: 'P2GS54',
+      schedules: [{date, startTime: '10:30:00', endTime: '10:50:00', status: 'PERFORMANCE_TIME'}],
+    },
+  ]);
+
+  return park;
+}
+
+/** Minutes between two ISO timestamps. */
+function spanMinutes(startTime: string, endTime: string): number {
+  return (new Date(endTime).getTime() - new Date(startTime).getTime()) / 60000;
+}
+
+describe('DLP showtimes', () => {
+  beforeEach(() => vi.restoreAllMocks());
+
+  it('derives the end time from the POI duration', async () => {
+    const live = await stubbedPark().getLiveData();
+
+    const parade = live.find((l) => l.id === 'P1GS21');
+    expect(parade?.showtimes).toHaveLength(1);
+    expect(spanMinutes(parade!.showtimes![0].startTime!, parade!.showtimes![0].endTime!)).toBe(30);
+  });
+
+  it('adds hours and minutes together', async () => {
+    const live = await stubbedPark().getLiveData();
+
+    const fireworks = live.find((l) => l.id === 'P1GS99');
+    expect(spanMinutes(fireworks!.showtimes![0].startTime!, fireworks!.showtimes![0].endTime!)).toBe(65);
+  });
+
+  it('keeps the feed end time when no duration is published', async () => {
+    const live = await stubbedPark().getLiveData();
+
+    const show = live.find((l) => l.id === 'P2GS54');
+    expect(spanMinutes(show!.showtimes![0].startTime!, show!.showtimes![0].endTime!)).toBe(20);
+  });
+
+  it('resolves durations without a preceding getEntities call', async () => {
+    // getLiveData is a standalone entry point; durations must not depend on
+    // entity building having run first.
+    const park = stubbedPark();
+    const live = await park.getLiveData();
+
+    const parade = live.find((l) => l.id === 'P1GS21');
+    expect(spanMinutes(parade!.showtimes![0].startTime!, parade!.showtimes![0].endTime!)).toBe(30);
+  });
+});
+
+/** Feeds one show with the given duration through the live-data path. */
+function parkWithDuration(duration: unknown): DisneylandParis {
+  const park = new DisneylandParis();
+
+  vi.spyOn(park as any, 'getPOIData').mockResolvedValue({
+    ThemePark: [{id: 'P1', name: 'Disneyland Park', type: 'ThemePark'}],
+    Entertainment: [{
+      id: 'P1GS21',
+      name: 'Disney Stars on Parade',
+      type: 'Entertainment',
+      subType: 'Parade',
+      location: {id: 'P1'},
+      duration,
+    }],
+  });
+  vi.spyOn(park as any, 'getWaitTimes').mockResolvedValue([]);
+  vi.spyOn(park as any, 'getPremierAccess').mockResolvedValue([]);
+  vi.spyOn(park as any, 'getVirtualQueueData').mockResolvedValue([]);
+  vi.spyOn(park as any, 'getScheduleForDate').mockImplementation(async (date) => [
+    {
+      id: 'P1GS21',
+      schedules: [{date, startTime: '17:30:00', endTime: '17:35:00', status: 'PERFORMANCE_TIME'}],
+    },
+  ]);
+
+  return park;
+}
+
+async function spanFor(duration: unknown): Promise<number> {
+  const live = await parkWithDuration(duration).getLiveData();
+  const show = live.find((l) => l.id === 'P1GS21');
+  return spanMinutes(show!.showtimes![0].startTime!, show!.showtimes![0].endTime!);
+}
+
+describe('DLP showtime duration parsing', () => {
+  beforeEach(() => vi.restoreAllMocks());
+
+  // The feed reports 17:30 to 17:35, so a 5 minute span means the duration was
+  // rejected and the feed's own end time survived.
+  it.each([
+    ['missing', undefined],
+    ['empty', {}],
+    ['null members', {hours: null, minutes: null}],
+    ['negative', {hours: 0, minutes: -20}],
+    ['not a number', {hours: 'soon', minutes: 'later'}],
+    ['longer than a day', {hours: 99999, minutes: 0}],
+  ])('falls back to the feed end time when the duration is %s', async (_label, duration) => {
+    expect(await spanFor(duration)).toBe(5);
+  });
+
+  it('coerces numeric strings instead of concatenating them', async () => {
+    expect(await spanFor({hours: '1', minutes: '30'})).toBe(90);
+    expect(await spanFor({minutes: '45'})).toBe(45);
+  });
+});
--- a/src/parks/dlp/__tests__/showtimes.test.ts
+++ b/src/parks/dlp/__tests__/showtimes.test.ts
@@ -1,4 +1,4 @@
-import {describe, it, expect, vi, beforeEach} from 'vitest';
+import {describe, it, expect, vi, afterEach} from 'vitest';
 import {DisneylandParis} from '../disneylandparis.js';
 
 /**
@@ -52,8 +52,9 @@ function stubbedPark(): DisneylandParis {
       schedules: [{date, startTime: '22:45:00', endTime: '22:45:00', status: 'PERFORMANCE_TIME'}],
     },
     {
+      // No duration published — the feed's zero-length slot survives.
       id: 'P2GS54',
-      schedules: [{date, startTime: '10:30:00', endTime: '10:50:00', status: 'PERFORMANCE_TIME'}],
+      schedules: [{date, startTime: '10:30:00', endTime: '10:30:00', status: 'PERFORMANCE_TIME'}],
     },
   ]);
 
@@ -66,7 +67,7 @@ function spanMinutes(startTime: string, endTime: string): number {
 }
 
 describe('DLP showtimes', () => {
-  beforeEach(() => vi.restoreAllMocks());
+  afterEach(() => vi.restoreAllMocks());
 
   it('derives the end time from the POI duration', async () => {
     const live = await stubbedPark().getLiveData();
@@ -87,17 +88,25 @@ describe('DLP showtimes', () => {
     const live = await stubbedPark().getLiveData();
 
     const show = live.find((l) => l.id === 'P2GS54');
-    expect(spanMinutes(show!.showtimes![0].startTime!, show!.showtimes![0].endTime!)).toBe(20);
+    expect(show?.showtimes).toHaveLength(1);
+    expect(spanMinutes(show!.showtimes![0].startTime!, show!.showtimes![0].endTime!)).toBe(0);
   });
 
-  it('resolves durations without a preceding getEntities call', async () => {
-    // getLiveData is a standalone entry point; durations must not depend on
-    // entity building having run first.
-    const park = stubbedPark();
-    const live = await park.getLiveData();
+  it('resolves the same durations with and without a preceding getEntities call', async () => {
+    // getLiveData is a standalone entry point; both call orders must agree.
+    const cold = stubbedPark();
+    const coldLive = await cold.getLiveData();
 
-    const parade = live.find((l) => l.id === 'P1GS21');
-    expect(spanMinutes(parade!.showtimes![0].startTime!, parade!.showtimes![0].endTime!)).toBe(30);
+    const warm = stubbedPark();
+    await warm.getEntities();
+    const warmLive = await warm.getLiveData();
+
+    for (const live of [coldLive, warmLive]) {
+      const parade = live.find((l) => l.id === 'P1GS21');
+      expect(spanMinutes(parade!.showtimes![0].startTime!, parade!.showtimes![0].endTime!)).toBe(30);
+    }
+    expect(warmLive.find((l) => l.id === 'P1GS21')?.showtimes)
+      .toEqual(coldLive.find((l) => l.id === 'P1GS21')?.showtimes);
   });
 });
 
@@ -136,7 +145,7 @@ async function spanFor(duration: unknown): Promise<number> {
 }
 
 describe('DLP showtime duration parsing', () => {
-  beforeEach(() => vi.restoreAllMocks());
+  afterEach(() => vi.restoreAllMocks());
 
   // The feed reports 17:30 to 17:35, so a 5 minute span means the duration was
   // rejected and the feed's own end time survived.
@@ -144,6 +153,7 @@ describe('DLP showtime duration parsing', () => {
     ['missing', undefined],
     ['empty', {}],
     ['null members', {hours: null, minutes: null}],
+    ['zero', {hours: 0, minutes: 0}],
     ['negative', {hours: 0, minutes: -20}],
     ['not a number', {hours: 'soon', minutes: 'later'}],
     ['longer than a day', {hours: 99999, minutes: 0}],
--- a/src/parks/dlp/__tests__/virtualQueue.test.ts
+++ b/src/parks/dlp/__tests__/virtualQueue.test.ts
@@ -130,6 +130,27 @@ describe('DLP virtual queue', () => {
     expect(row.queue?.RETURN_TIME).toBeUndefined();
   });
 
+  // A dormant configuration row: Disney leaves these in the wave list with
+  // both dates nulled, so they carry no evidence of having run.
+  const dormant = {waveId: 'w0', name: 'EMT', openAt: null, closedAt: null, status: 'FINISHED'};
+
+  it('does not read a dateless FINISHED wave as the day having started', async () => {
+    const row = await liveRowOf(
+      stubbedPark([morning('CLOSED'), afternoon('CLOSED'), dormant]),
+    );
+    expect(row).toBeDefined();
+    expect(row.status).toBe('CLOSED');
+    expect(row.queue?.RETURN_TIME).toBeUndefined();
+  });
+
+  it('still counts a FINISHED wave that kept its dates', async () => {
+    const row = await liveRowOf(
+      stubbedPark([morning('FINISHED'), afternoon('CLOSED'), dormant]),
+    );
+    expect(row.queue?.RETURN_TIME).toEqual({state: 'TEMP_FULL', returnStart: null, returnEnd: null});
+    expect(row.status).toBe('OPERATING');
+  });
+
   it('finishes once the last wave is full', async () => {
     const row = await liveRowOf(
       stubbedPark([morning('FINISHED'), afternoon('FULL')], {nextWaveId: 'w2'}),
--- a/src/parks/dlp/__tests__/virtualQueue.test.ts
+++ b/src/parks/dlp/__tests__/virtualQueue.test.ts
@@ -118,6 +118,9 @@ describe('DLP virtual queue', () => {
       stubbedPark([morning('FULL'), afternoon('CLOSED')], {nextWaveId: 'w1'}),
     );
     expect(row.queue?.RETURN_TIME).toEqual({state: 'TEMP_FULL', returnStart: null, returnEnd: null});
+    // "Come back later" says the experience is running, so the seeded CLOSED
+    // would contradict the queue it is published next to.
+    expect(row.status).toBe('OPERATING');
   });
 
   it('publishes no return time before the first wave of the day has opened', async () => {
@@ -132,11 +135,14 @@ describe('DLP virtual queue', () => {
       stubbedPark([morning('FINISHED'), afternoon('FULL')], {nextWaveId: 'w2'}),
     );
     expect(row.queue?.RETURN_TIME).toEqual({state: 'FINISHED', returnStart: null, returnEnd: null});
+    // Nothing left to open, so the seeded CLOSED is the honest answer.
+    expect(row.status).toBe('CLOSED');
   });
 
   it('finishes when every wave is done', async () => {
     const row = await liveRowOf(stubbedPark([morning('FINISHED'), afternoon('FINISHED')]));
     expect(row.queue?.RETURN_TIME).toEqual({state: 'FINISHED', returnStart: null, returnEnd: null});
+    expect(row.status).toBe('CLOSED');
   });
 
   it('never claims availability for a status it does not recognise', async () => {
--- a/src/parks/dlp/__tests__/virtualQueue.test.ts
+++ b/src/parks/dlp/__tests__/virtualQueue.test.ts
@@ -1,14 +1,18 @@
-import {describe, it, expect, vi, beforeEach} from 'vitest';
+import {describe, it, expect, vi, afterEach} from 'vitest';
 import {DisneylandParis} from '../disneylandparis.js';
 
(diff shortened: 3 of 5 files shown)


BeyondVertical:
Two follow-ups, both driven by the virtual-queue feed now being configured locally — until today `getVirtualQueueData()` always returned an empty array, so this branch's VQ path had never actually been exercised against live data.

**1. A full queue was published as bookable.** Every wave that was not FINISHED emitted `AVAILABLE` together with the wave's own window. Only an `OPEN` wave takes bookings: `CLOSED` is scheduled to open later and `FULL` is that wave's allocation gone. Measured at 12:04, four of the queues this branch surfaces read FULL with zero of six group sizes available — and all four were being advertised as available. They now report `TEMP_FULL` while a CLOSED wave is still to come, `FINISHED` once none is, and only an OPEN wave publishes a window. The mapping was also hardened: a wave list that is not an array, a null wave and a non-string status each used to throw out of `buildLiveData`.

**2. Two more Heroic Encounters.** `P2MG33` (Spider-Man) and `P2MG43` (MARVEL Super Hero) carry `Hide from the Service` but run live virtual queues. Sampling the feed every three minutes from 12:56 to 14:14 across the 14:00 wave change:

| | 13:59 | 14:02 | 14:05 |
|---|---|---|---|
| P2MG33 | CLOSED | FULL 0/6 | FULL 0/6 |
| P2MG43 | CLOSED | **OPEN 6/6** | **OPEN 6/6** |
| P2MG73 *(published)* | CLOSED | OPEN 6/6 | FULL 0/6 |
| P1M116 *(published)* | CLOSED | FULL 0/6 | FULL 0/6 |

P2MG43 was directly observed bookable at every group size. P2MG33 shows the same trace as P1M116, which this branch already publishes — dead configuration would have stayed CLOSED or gone FINISHED rather than transitioning to FULL at the scheduled opening.

The control in the test is the retired twin: `P2MG59` carries the same name as the published `P2MG73` and the same flag, but no virtual queue, and stays filtered.

Entities 149 → 151.

---

**On merging:** `main` has been merged in (#292's conflict in `VISIBILITY_EXCEPTIONS` and `visibility.test.ts` is resolved keep-both). #292's comment there said its two ids "are the only members of that flag we surface"; `P2MG31`, `P1MG21` and `P1MG05` on this branch carry the same flag, so that sentence is reworded in the merge.

One thing git will not catch: `getPOIData`'s `cacheVersion` goes 2 → 3 both here and in #295, and those two identical edits merge silently, leaving 3. Whichever lands second needs 4 by hand.

cubehouse (changes requested):
Reviewed this in depth, including running the tests against pre-PR source and driving the virtual-queue path through an exhaustive status matrix. The showtime work is a clear improvement and the meet & greet evidence is solid. Two things need fixing before it lands, both in the VQ path.

First, a correction to my own earlier read: this branch is **not** conflicting. It carries a merge commit from `main`, `git merge-tree main pr293` returns a clean tree, and #292's changes are already in it. No rebase needed.

## Blocking

### 1. `status: CLOSED` published alongside `RETURN_TIME: AVAILABLE`

`disneylandparis.ts:922`. `getOrCreate()` seeds a new row with `status: 'CLOSED'` and the VQ block never updates it. For a meet & greet whose only live signal is the virtual queue, the published row contradicts itself.

Verified by driving a 2-wave x 4-status x 3-`nextWaveId` matrix through the real `getLiveData()`. All 48 combinations came back `liveStatus=CLOSED`, including:

```
[OPEN,CLOSED] next=- => state=AVAILABLE liveStatus=CLOSED start=2026-08-12T01:00:00+02:00
```

This is newly reachable and worth calling out plainly: on `main` the entire VQ block was dead code. All 7 `queueContentId`s in the live feed (`P1M115`, `P1M116`, `P1MG19`, `P2MG33`, `P2MG43`, `P2MG66`, `P2MG73`) are `Character Experience - Meet & Greet`, which `mapEntityType` returned `undefined` for, so `getOrCreate` returned `undefined` and every one was skipped. Adding the subtype to `SHOW_SUBTYPES` is what switches this path on for the first time. That raises the bar on the whole block rather than just this line.

Setting `ld.status = 'OPERATING'` when the state is `AVAILABLE` is the fix. Compare `efteling.ts:484`, where `ld` already carries a real status from the wait feed; DLP's does not.

### 2. `TEMP_FULL` published overnight, roughly 10 hours a day

`disneylandparis.ts:935`. Checked against the live feed at 00:01 Paris: all six enabled queues had every dated wave at `CLOSED`, first opening 09:45, and all six published:

```
P1M115  status=CLOSED  queue={"RETURN_TIME":{"state":"TEMP_FULL","returnStart":null,"returnEnd":null}}
P1M116 / P2MG33 / P2MG43 / P2MG66 / P2MG73  same
```

`TEMP_FULL` means "full, come back later". At midnight nothing has opened yet, so this trades one wrong state for another. The test at `virtualQueue.test.ts:100` pins the behaviour, so it is deliberate, but I do not think it is right. Leaving `RETURN_TIME` unset until a wave has actually been `OPEN` or `FULL` at least once that day is the honest option.

### 3. A stale `nextWaveId` hides a live OPEN wave

`disneylandparis.ts:920`. The code trusts `activeWave` absolutely and never asks whether any wave is open. 8 of the 48 states publish `FINISHED` or `TEMP_FULL` while an `OPEN` wave exists:

```
[FINISHED,OPEN] next=w1 => FINISHED    <- OPEN wave exists
[FULL,OPEN]     next=w1 => FINISHED    <- OPEN wave exists
[OPEN,FULL]     next=w2 => FINISHED    <- OPEN wave exists
[OPEN,CLOSED]   next=w2 => TEMP_FULL   <- OPEN wave exists
```

The realistic trigger is `nextWaveId` lagging a transition. Pre-PR this shape published `AVAILABLE` with a wrong window; post-PR it publishes a hard `FINISHED`, which is a stickier failure. Checking for an OPEN wave first fixes it:

```ts
const activeWave =
  waves.find((w) => waveStatus(w) === 'OPEN') ||
  (q.nextWaveId && waves.find((w) => w?.waveId === q.nextWaveId)) || pending[0];
```

Worth noting that change passes all 30 tests unmodified, which is itself the point: the suite cannot tell the two behaviours apart.

## Answering my own earlier question

`waves.some(CLOSED)` vs `pending.some(CLOSED)` is **not** a bug. `pending` excludes exactly the `FINISHED` waves and `FINISHED !== 'CLOSED'`, so the predicates cannot differ. Brute-forced all 64 three-wave combinations: 0 differ. Redundant, not wrong, so ignore my earlier note. Related evidence for the underlying worry: at 23:52 Paris all 19 waves across 7 queues read `FINISHED` with `openAt`/`closedAt` nulled, so DLP does retire passed waves.

## Test gaps

- **3 of the 5 new `VISIBILITY_EXCEPTIONS` are untested.** Deleting `P2MG31`, `P1MG21` and `P1MG05` from the set leaves 30/30 green. No fixture covers them.
- **The single largest change in the PR has no assertion.** Roughly 16 unflagged meet & greets become `SHOW` entities and nothing tests it.
- `showtimes.test.ts:69` is a textual duplicate of the test at `:46`: same stub, same call, same assertion, same expected `30`. If it is meant to guard the `showDurationMap` removal it needs a variant that calls `getEntities()` first and one that does not, then asserts they agree.
- `showtimes.test.ts:151` is tautological. Deleting the whole `if (!Number.isFinite(hours) || !Number.isFinite(minutes)) return 0;` line leaves 30/30 passing, because `Math.round(NaN*60+NaN)` is `NaN` and `NaN > 0` is already false. The guard is dead and its test proves nothing.
- `virtualQueue.test.ts` fixtures use `subType: 'Character Experience - Meet & Greet'`, so removing that one line from `SHOW_SUBTYPES` fails 8 of the 11 VQ tests for a reason unrelated to virtual queues. Use `'Stage Show'`.
- The 5 `it.each` "never publishes a bookable window" cases and `publishes no queue at all while the queue is disabled` are vacuous: they assert `expect(queue?.state).not.toBe('AVAILABLE')` where `queue` is `undefined` when no live row exists at all. They pass even when the code produces nothing.
- No test asserts the `status` field, which is exactly where finding 1 hides.
- Not tested: `{hours: 0, minutes: 0}`, which is the most common real shape (15 of the 44 live POIs with a `duration` have it).
- `beforeEach(() => vi.restoreAllMocks())` runs before the mocks are created, since they are made inside the test body. Should be `afterEach`, in all three files.

## Two claims in the description that overstate the result

Neither affects the code, both worth correcting so the commit history stays accurate.

- **The `(minutes||0) + ((hours||0)*60)` concatenation bug is unreachable**, on `main` and on the live feed. `main`'s `fetchPOI` never requested `duration`, so `showDurationMap` was always empty. Today all 44 POIs with a duration have `hours: 0` and numeric values; old expression vs new, 0 differ. The fix is worth keeping as hardening, but the actual fix in that hunk is adding `duration` to the query.
- **The end-time fix lands for 8 of 30 shows.** 22 have no `duration` in POI at all, including Stitch Live!, Animation Academy, Mickey's PhilharMagic and every meet & greet, so they still publish zero-length windows. Not a defect, but the title promises more than it delivers, and the fallback test at `showtimes.test.ts:64` uses a synthetic `10:30 -> 10:50` that never occurs in production.

## On the five exceptions

I verified the three meet & greets independently over 30 days of schedule feed: `P2MG31` 30/30, `P1MG05` 30/30, `P1MG21` 29/30 days with performances, all busier than already-published controls (`P1GS21` 26/30, `P1GS93` 27/30). Those three are well justified.

`P2MG33` and `P2MG43` are the weak ones. They get zero schedule rows over 60 days, so they publish as shows with no schedules and, outside VQ hours, no live row. They lean entirely on the path carrying finding 1. Worth fixing that first and then revisiting whether they earn their place.

Nothing blocking beyond the three above. The `from && until` requirement for `AVAILABLE` is correct rather than over-strict, per `src/virtualQueue/validator.ts:64-73`, and the cache invalidation is sound.

One merge note: this and #295 both take `getPOIData` to `cacheVersion: 3`. Git resolves the identical edit silently, so whichever lands second needs `4`.

BeyondVertical:
All three blocking findings are fixed in 3a07e57d.

**1. Status vs. AVAILABLE** — a bookable window now sets `status: 'OPERATING'` on the row. Your 48-state matrix case (`[OPEN,CLOSED] next=-`) now comes back `state=AVAILABLE liveStatus=OPERATING`, and the AVAILABLE tests assert the status field directly.

**2. Overnight TEMP_FULL** — no `RETURN_TIME` is published until a wave has actually been in play (OPEN, FULL or FINISHED). Overnight the six queues now emit a plain CLOSED row with no queue block. One deliberate side effect: an unrecognised wave status paired with CLOSED waves also publishes nothing now, rather than the guessed TEMP_FULL the old test pinned.

**3. Stale `nextWaveId`** — a live OPEN wave wins over `nextWaveId`, per your snippet. Your four failing states are in the suite as an `it.each` matrix, and the point about the old suite not telling the behaviours apart holds: 11 of the rewritten VQ tests fail against the pre-fix source.

On the test gaps, all taken:

- The `it.each` malformed-payload cases and the disabled-queue test now assert on the live row itself, distinguishing "no row" from "row without RETURN_TIME" instead of the vacuous `queue?.state` chain.
- VQ fixtures use `Stage Show`, so the suite no longer depends on the meet & greet subtype.
- `P2MG31`/`P1MG21`/`P1MG05` and the unflagged meet & greet → SHOW mapping are covered; deleting any of the three from `VISIBILITY_EXCEPTIONS` now goes red.
- The duplicated showtimes test became a cold/warm agreement check (`getLiveData()` alone vs. after `getEntities()`).
- The dead `Number.isFinite` guard is removed — you're right that the range check already catches NaN and Infinity. The `not a number` case stays as a behaviour test of the fallback, and `{hours: 0, minutes: 0}` is added.
- The fallback fixture uses the production zero-length shape now; the synthetic 5-minute window survives only in the parsing suite, where it is the detector.
- `afterEach` in all three files.

Description corrected as well: the concatenation claim is reworded as hardening of an unreachable-on-`main` path, and the window count now reads 8 of 30.

On `P2MG33`/`P2MG43`: I'd keep them. With finding 1 fixed they publish OPERATING plus a window during open waves, a plain CLOSED row otherwise, and nothing misleading overnight — and P2MG43 was directly observed bookable at every group size. If you'd rather they wait until schedule rows appear, I'll split them out.

The `cacheVersion` note stands: whichever of this and #295 lands second goes to 4.

cubehouse:
The three virtual-queue blockers are fixed correctly. `:947-950` sets `status: 'OPERATING'` alongside an `AVAILABLE` window, the `dayStarted` gate at `:932-935` stops `TEMP_FULL` shipping overnight, and the OPEN-wave preference at `:918-921` stops a stale `nextWaveId` hiding a bookable window. The test hardening is real too: the vacuous optional-chain assertions are gone, the fixture no longer leans on the meet & greet subtype, and the dead finiteness guard in `showDurationMinutes` is deleted rather than tested.

Not approving yet, because the part of this PR that was never the virtual queue needs a decision. Measured against the live feed on 2026-08-14 at 18:35 Paris, both parks open on published hours 09:30 to 22:00.

## 8 of the 22 new meet & greet entities carry no data at all

Adding `Character Experience - Meet & Greet` to `SHOW_SUBTYPES` publishes 22 entities. Live data and schedules for each, sampled with the parks open:

```
P1MG21 An Encounter with Captain Hook               ----  sched
P2MG32 Hero Training Center                         ----  -----
P2MG43 MARVEL Super Hero Heroic Encounter           ----  -----
P2MG72 Meet Bo Peep                                 live  sched
P1MG05 Meet Donald Duck or his friends              live  sched
P2MG31 Meet Goofy, the Movie Director               ----  sched
P1MG33 Meet Max                                     ----  sched
P1MG11 Meet Merida                                  ----  sched
P2MG71 Meet Mickey                                  live  sched
P1M115 Meet Mickey Mouse                            ----  -----
P1MG86 Meet Minnie or her friends across Europe     live  sched
P1M133 Meet Sofia the First                         ----  sched
P2MG15 Meet Woody                                   live  sched
P1MG55 Meet a Character from Aladdin                ----  sched
P1MG36 Meet a Character from The Lion King          ----  sched
P2MG73 Meet a Toy Story Character                   ----  -----
P1MG23 Meet the Characters of Alice in Wonderland   ----  sched
P1MG65 Mysterious Meeting                           live  sched
P1M116 Princess Pavilion: A Royal Invitation        ----  -----
P2MG66 Rencontre Royale                             ----  -----
P2MG33 Spider-Man Heroic Encounter                  ----  -----
P1M117 Starport                                     ----  -----
```

14 of the 22 are clearly worth publishing. The other 8 land on the wiki as bare records with neither live data nor a single schedule day. That is the `P2MG33` / `P2MG43` concern from the first review, doubled once the whole set is measured rather than just the exception list.

Six of the eight are virtual-queue ids, so their only possible data source is the VQ block, which brings us to:

## The virtual-queue path is inert right now

All seven queues read `enabled: false` at 18:35 with the parks open, every wave `FINISHED`, and `nextWaveId` undefined on all seven:

```
P1M115 enabled=false next=undefined  2 waves, both FINISHED
P1M116 enabled=false next=undefined  3 waves, all FINISHED
P1MG19 enabled=false next=undefined  2 waves, both FINISHED   (no published POI record)
P2MG33 enabled=false next=undefined  3 waves, all FINISHED
P2MG43 enabled=false next=undefined  3 waves, all FINISHED
P2MG66 enabled=false next=undefined  3 waves, all FINISHED
P2MG73 enabled=false next=undefined  2 waves, all FINISHED
```

`:912` skips on `enabled === false` before `getOrCreate` is ever called, so none of the new behaviour is exercised in production today. It is test-only until Disney turns a queue back on. That is not an argument against the fixes, they are right and they should land. It does mean the six VQ meet & greets have no data source at all in the meantime.

Worth noting the wave list is historical rather than per-day. Several of those wave ids decode to 2022 and 2025 timestamps, so `dayStarted` reading `FINISHED` as "a wave has been in play" will be true from stale waves the moment a queue is re-enabled. The `FINISHED` branch is the honest answer there, so I think this is fine, but it is not the "today's waves" model the comment at `:906-911` describes.

## Still open from the first review

`:952` leaves `status` on the seeded `CLOSED` in the `TEMP_FULL` branch. `TEMP_FULL` means "full, come back later", which implies the experience is running, so this is the same self-contradiction the `AVAILABLE` branch just stopped publishing, one branch over. Latent while every queue is disabled, live the moment one is enabled and fills.

## The end-time work lands narrower than the title

Of 107 showtime slots emitted right now, 24 carry a real duration and 83 are still zero-length. 19 live rows carry showtimes. The fix is correct and worth having, the title just promises more of the estate than the POI `duration` field can currently cover.

## What I would do

Split it. The virtual-queue fixes, the showtime end times and the schedule-backed exceptions are all good and I would take them today. Hold the `SHOW_SUBTYPES` line, or pair it with a filter that drops a meet & greet with no schedule rows and no live source, so the 14 with data ship and the 8 without do not.

## Merge notes

- `getPOIData` is on `cacheVersion: 3` here and #295 has already moved to 4. Whichever lands second needs 4 or higher.
- This branch conflicts with #294, #295, #296 and #297, and with none of #298 or #299.

BeyondVertical:
Taken the filter rather than the split, plus the `status` leftover and the cacheVersion. Rebased onto current `main`, so #295 is in and this is mergeable again.

## Why the filter and not the split

Holding the `SHOW_SUBTYPES` line does not leave the 14 with data and drop the 8 — it drops all 22. `mapEntityType` at `:710-720` returns `undefined` for any `Entertainment` record whose subtype is not in the set, so without that line no meet & greet reaches `getOrCreate` at all, and the five `VISIBILITY_EXCEPTIONS` entries go inert with it: all five are meet & greets, so they would bypass the hide rules and then fail the type mapping anyway. A split would ship the showtime end times and the virtual-queue fixes and nothing else.

So the gate is on the data instead. A meet & greet publishes when it has a performance somewhere in the window `buildSchedules` covers, or a virtual queue Disney has switched on.

Deliberately the same `SCHEDULE_DAYS` window on both sides, so an entity we publish is one we can attach hours to — otherwise this creates exactly the orphan class #299 exists to detect. `buildEntityList` now goes through `getEmittablePOIEntities` as well, which the live and schedule paths already used, so all three share one definition.

An outage must widen the set, not empty it: the sweep returns null when no day in the window answered with a single row, and null means publish all of them. Same trap #296 closed for schedules; a cached empty answer would otherwise hold for 12 hours.

## Measured

At first, with the parks shut and all seven queues reading `enabled: false`, the split came out exactly as your table: 22 meet & greets in POI, 14 published, 8 held back — the same 8, id for id.

Then it moved on its own, which is the part worth reporting:

```
22 in POI   20 published   2 held back
held: P1M117 Starport, P2MG32 Hero Training Center
```

Six of the seven queues had been switched on for the coming day, so the six virtual-queue meet & greets came back with no code change. That is the property I wanted out of a filter rather than a second hardcoded list, and it also revises one of your findings below.

Zero orphans in both directions, on both measurements. Entities 151, harness 4/4.

## Your remaining finding: status on the TEMP_FULL branch

Fixed. `TEMP_FULL` sets `OPERATING` too — "come back later" says the experience is running. `FINISHED` keeps the seeded `CLOSED`: nothing is left to open. Both branches assert the status field now.

## The virtual-queue path is not inert, and the FINISHED branch is not honest

Two corrections to the second review, both from the live feed rather than reasoning.

**The queues do get switched on.** You measured at 18:35 with all seven `enabled: false` and read that as "test-only until Disney turns a queue back on". Measured again after midnight, six of the seven were enabled with today's waves dated and scheduled. It looks like they are disabled while the day's allocation is spent and re-enabled for the next day, so 18:35 caught the trough rather than a dormant configuration.

**And that exposed a real defect.** You predicted the mechanism — "`dayStarted` reading `FINISHED` as 'a wave has been in play' will be true from stale waves the moment a queue is re-enabled" — but concluded "the `FINISHED` branch is the honest answer there, so I think this is fine". It is not the FINISHED branch that runs. With today's waves still `CLOSED` and one stale row present, the code takes the `TEMP_FULL` branch, because `waves.some(CLOSED)` is true. Three queues published "temporarily full, come back later" in the middle of the night — and, after the fix above, `OPERATING` with it:

```
P2MG33  CLOSED   open=<today> 09:45  close=<today> 13:30
        CLOSED   open=<today> 14:00  close=<today> 17:50
        FINISHED open=null           close=null      <- wave id decodes to 2022
```

The dormant rows are distinguishable after all: they carry no dates. A `FINISHED` wave now counts as history only while it still holds its `openAt`, which also covers your other observation — Disney nulls the dates of the day's real waves once they have passed, so those stop counting at the same moment they stop being today's. `OPEN` and `FULL` are unchanged; those say the queue is distributing whatever the dates read.

After the change all six enabled queues publish a plain `CLOSED` row with no queue block, which is the honest state before a wave has opened. A `FINISHED` wave that kept its dates still counts, so a queue whose morning wave has passed and whose afternoon wave is still to come keeps reporting `TEMP_FULL`. Both are under test, and each fix has a test that fails when the line is reverted.

## Merge note

`getPOIData` is on `cacheVersion: 5` — #295 took 4 and the query text differs between the two, so they need separate keys.

## Cost

One thing to weigh: `getEntities` now runs the schedule sweep, so on a cold cache it takes about 34s instead of 1s. It is the same 60 days `buildSchedules` fetches and `getScheduleForDate` caches for 24h, so a full pass pays it once rather than twice and a warm `getEntities` is 31ms. But a consumer that only ever calls `getEntities` does pay it. Happy to narrow the window if you would rather — 30 days would cover every meet & greet that is publishing today, though it would reintroduce the drift against the 60-day schedule output.

Suite 1882 passing across 81 files, `tsc --noEmit` clean.

cubehouse (changes requested):
The filter was the right call over the split, and the correction you pushed back on the second review is the more useful half of this round: I read the dateless `FINISHED` rows as harmless and they were not, and the mechanism you found is the one that runs.

Measured on a tree with current `main` merged in, parks open, 09:00 to 09:30 Paris.

## What holds up

**The data gate does what it claims.** 131 to 151 entities, and every one of the 20 new meet & greets carries live data, schedule rows, or both:

```
P1M115 live ---- Meet Mickey Mouse            P1MG55 live sched Meet a Character from Aladdin
P1M116 live ---- Princess Pavilion            P1MG65 live sched Mysterious Meeting
P1M133 live sched Meet Sofia the First        P1MG86 live sched Meet Minnie or her friends
P1MG05 live sched Meet Donald Duck            P2MG15 live sched Meet Woody
P1MG11 live sched Meet Merida                 P2MG31 live sched Meet Goofy
P1MG21 live sched An Encounter with Hook      P2MG33 live ---- Spider-Man Heroic Encounter
P1MG23 ---- sched Alice in Wonderland         P2MG43 live ---- MARVEL Heroic Encounter
P1MG33 live sched Meet Max                    P2MG66 live ---- Rencontre Royale
P1MG36 ---- sched The Lion King               P2MG71 live sched Meet Mickey
                                              P2MG72 live sched Meet Bo Peep
                                              P2MG73 live ---- Meet a Toy Story Character
```

Zero orphans in both directions, so #299's check stays quiet. Showtime slots 90 to 179, real windows 0 to 41.

**The virtual-queue path is live, and the `dayStarted` fix is doing exactly what you designed it to.** At 08:59, before opening:

```
P2MG66 enabled=true
   FULL     08:45 -> 09:30
   CLOSED   09:45 -> 13:30
   CLOSED   14:00 -> 19:45
   => OPERATING / TEMP_FULL

P1M116, P2MG33, P2MG43   two dated CLOSED waves + one FINISHED with openAt=null
   => plain CLOSED row, no queue block
```

A genuinely full early wave publishes "come back later"; the dormant 2022-era rows are ignored on all three queues that carry one. That is the right answer in both cases, arrived at from the feed rather than from the wave order.

**Merging `main` is mechanical.** You are two commits behind, #298 and #299, and #298 edits this same file. I merged it locally: clean, `tsc --noEmit` clean, 188 DLP tests green, harness 4/4. Push the merge so CI runs against it, but there is nothing to untangle.

## Blocking: the outage sentinel cannot be cached

`disneylandparis.ts:871`. `getScheduledActivityIds` returns `null` when no day answered, and `@cache` cannot store that. `CacheLib.wrap` reads through `CacheLib.get`, and `cache.ts:435` treats a cached `null` as a miss:

```
null return  -> fn invoked 3 times over 3 wrap() calls (cached would be 1)
array return -> fn invoked 1 time
```

So the 12h TTL on that method is inert on exactly the path it was written for. During a schedule-feed outage the 60-day sweep re-runs on every call, and `getEmittablePOIEntities` is reached from all three public entry points, so one poll cycle costs 180 failed upstream requests instead of 60, forever, with no backoff. The healthy path is fine; it is only the outage that never settles, which is the worst time for it.

Returning something cacheable fixes it without touching the semantics:

```ts
private async getScheduledActivityIds(): Promise<{answered: boolean; ids: string[]}> {
  ...
  return {answered, ids: [...ids]};
}
```

and `if (!scheduled.answered) return null;` at the call site keeps "null means publish all" where it already is.

## Notes

**`dayStarted` assumes a passed wave loses its dates promptly.** `:1107` counts any `FINISHED` wave that still holds `openAt` as evidence the day has started. True on today's feed, where the day's waves carry today's dates and the dormant rows carry none. If the feed ever holds yesterday's dated waves into the small hours, `TEMP_FULL` plus `OPERATING` comes straight back at 3am. Requiring `openAt` to fall on today's park date closes that for one line and costs nothing, since you already have `todayStr` in scope.

**Cold `getEntities` is 75s here**, against 255ms on `main`; warm is 351ms against 255ms. Higher than the 34s you measured, so it varies with the upstream. It is the same 60 days `buildSchedules` needs and a full pass pays it once, so I am not asking you to change it, but it is worth stating in the description as a real number rather than a footnote, because a consumer that only calls `getEntities` now pays a schedule sweep it never wanted.

**The 20 new entities have no coordinates.** The harness warning goes `SHOW:20` to `SHOW:40`. Disney publishes no location for meet & greets, so there is nothing to fix in this PR, but it doubles the count on that line and I would rather it was said out loud than discovered later.

Fix the sentinel and push the `main` merge and I will take it.

BeyondVertical:
Sentinel fixed, `main` merged, and the `dayStarted` note taken as well. Two commits on top of the merge, one per finding.

## The outage sentinel

You are right, and it is worse than inert — it is inert in the direction that costs the most.

Reproduced it as a call count rather than by reading `cache.ts`, with `getScheduleForDate` stubbed so every fetch is visible. Three calls to the sweep against a dead feed:

```
null return        180 fetches   (60 per call, cached never)
{answered, ids}     60 fetches   (60 once, then the cache answers)
```

Your suggested shape, verbatim, with `if (!scheduled.answered) return null;` at the call site so "null means publish all" stays exactly where it was. `CacheLib.get('DisneylandParis:dlp:getScheduledActivityIds')` now comes back populated on both paths.

The three tests live in their own file, because the sweep's fixed cache key is shared within a file and I did not want a later test to inherit an earlier one's answer. The counting one fails against the previous code with `expected 180 to be 60`.

## `dayStarted` and yesterday's dated waves

Taken. A `FINISHED` wave now has to have opened on the current park date, using the `todayStr` already in scope — one line, plus the helper that formats a wave's `openAt` into the park's date to compare it.

Worth saying what the guard is now layered against, because the two cases are different:

- a **dateless** `FINISHED` row is the dormant configuration Disney leaves in the list (the 2022-era ones);
- a **yesterday-dated** `FINISHED` row is the one you predicted, which the feed does not produce today.

The second is untested against reality by definition, so the test fixture builds it explicitly rather than waiting for the feed to do it. Reverting the line turns that test red with `expected 'OPERATING' to be 'CLOSED'`.

While doing it I moved the virtual-queue fixtures off their hardcoded `2026-08-12` and onto the park's current date. Pinned dates would have made the whole file stop exercising `dayStarted` the day after it was written — the tests passed only because every wave was equally stale.

## Merge

Clean, as you said. `main` had moved a little further by the time I got to it — #303 and #307 as well as #298 and #299 — and it is still mechanical. `tsc --noEmit` clean, DLP 192 tests, full suite 1945 across 88 files, harness 4/4.

One note on the suite: `traceContextPropagation.test.ts` failed once on a full run here and passed on its own and on the next full run. That is the live-endpoint flake you flagged in #302, not this branch.

## Measured after the merge, parks open

```
22 meet & greets visible in POI   20 published   2 held
held: P2MG32 Hero Training Center, P1M117 Starport
orphans: live 0, schedules 0
published meet & greets with neither live data nor a schedule: 0
```

Same two held back, id for id, as your table.

Virtual queues, six of seven enabled, mid-morning:

```
P1M115  FULL 09:45 -> CLOSED 14:00                     OPERATING / TEMP_FULL
P1M116  FULL 09:45 -> CLOSED 14:00 + FINISHED null     OPERATING / TEMP_FULL
P2MG33  FULL 09:45 -> CLOSED 14:00 + FINISHED null     OPERATING / TEMP_FULL
P2MG43  FULL 09:45 -> CLOSED 14:00 + FINISHED null     OPERATING / TEMP_FULL
P2MG66  FULL 09:45 -> CLOSED 14:00 + FINISHED null     OPERATING / TEMP_FULL
P2MG73  FULL 09:45 -> CLOSED 14:00                     OPERATING / TEMP_FULL
```

The morning allocation is genuinely gone and an afternoon wave is scheduled, so "come back later" is the right answer on all six — the same queues that published a plain `CLOSED` row before opening. The dormant rows are along for the ride on three of them and change nothing.

Showtimes: 29 live rows, 177 slots, 41 with a real window.

## The description

Both of your notes are in it now as their own section rather than as a footnote: the cold `getEntities()` cost with the real number (38s here against ~1s on `main`, warm 8ms, and I quote your 75s as the upper end I have seen reported), and the coordinates line going `SHOW:20` to `SHOW:40`. The stale figures from the first round went with it — entity counts, `cacheVersion`, and a test count that predated most of the tests.

I left the window at 60 days. Narrowing to 30 would cut the cold cost but reintroduce the drift against the 60-day schedule output, which is the orphan class #299 exists to catch. Say the word if you would rather have the cost than the invariant.

BeyondVertical:
`main` moved again while this was in flight — #302, #308 and #309 — so I merged it a second time and re-measured both columns back to back rather than leaving figures that were true against the older base.

```
                      main (1c8a739d)   this branch
entities                    127             147
SHOW                         23              43
showtime slots               90             177
  with a real window          0              41
```

#308 takes four building records out of `ATTRACTION` on both sides, which is why the totals sit lower than the 131 → 151 you measured. The meet & greet split is unchanged: 22 visible in POI, 20 published, the same two held back. Zero orphans in both directions, cold `getEntities()` 43s against ~1s, warm 7ms.

Merge clean, `tsc --noEmit` clean, DLP 202 tests across 10 files, full suite 1955 across 89 files, harness 4/4. The description carries these numbers now.

#309 also settles the flake I mentioned above — the full suite has been green on every run since.

cubehouse (approved):
Approving. Both findings are fixed properly, and the second one you took further than I asked.

## The sentinel

`{answered, ids}` is the right shape and the flag stays at the call site, so "publish all of them" reads the same as before. `scheduleSweepCache.test.ts` is the part that makes it stick: counting calls into `getScheduleForDate` tests the thing that actually costs something, rather than asserting on the return value and hoping. Three calls, one sweep, on both the healthy and the dead-feed paths.

## The wave date

I flagged this as cheap hardening against a hypothetical and you closed it against the park's clock instead. `openedToday()` is the correct comparison — the wave's `openAt` in park time against `todayStr` — and it keeps the two states that were already right: `OPEN` and `FULL` still say the queue is distributing whatever the dates read, and a `FINISHED` wave dated today still holds `TEMP_FULL` while a later wave is scheduled.

## Verified

Merged current `main` in locally and re-ran everything:

```
tsc --noEmit    clean
vitest          91 files / 1973 tests
npm run dev     4/4, 146 entities
                ATTRACTION 51, SHOW 42, RESTAURANT 50
                no orphans, no duplicate ids
```

The duplicate-id gate from #310 is in that tree and stays quiet, which is the answer to the thing I could not check on the first two rounds.

One difference from your figures worth naming, because it is the feature working rather than a discrepancy: you measured 147 entities and 43 shows, I got 146 and 42. One meet & greet moved out of the published set between the two runs. That is the data gate doing its job — the set is meant to follow the feed, and it does.

## One optional nit, not blocking

`getScheduledActivityIds` changed return shape but its `@cache` entry has no `cacheVersion`. The key is new in this PR so nothing in production holds an old value, and a stale array reads `.answered` as `undefined` and falls through to "publish all", which is the safe direction. The only exposure is a developer who ran an earlier build of this branch locally, for up to 12h. Adding `cacheVersion: 2` alongside the key would close it; equally fine to leave.

Thanks for three rounds of this. The meet & greet evidence, the virtual-queue work and the two corrections you pushed back on me for — the dormant dateless waves, and the queues not being inert — were all right, and the PR is better for each of them.

PR #294, opened 2026-08-11
feat(dlp): publish dining status and today's opening hours

BeyondVertical:
## Summary

Restaurants had no live data at all. Their hours were already fetched for `buildSchedules`, so a consumer could see a 60-day plan without being able to tell whether a venue is open right now.

| | before | after |
|---|---|---|
| Restaurant live rows | 0 | **31** of 50 |
| …carrying `operatingHours` | – | **30**, one slot each |
| …carrying a `queue` | – | **0** |
| DLP tests | 2 | **26** |

Entities, attractions, shows and schedules are untouched. No new HTTP request — the hours come from the activity-schedule payload the showtimes path already fetches for today. (Originally they were read from the POI blob's `schedules`, which only ever carry the fetch day; review caught that going dark from local midnight until the 12h POI cache rolled.)

```json
{
  "id": "P1AR06",
  "status": "OPERATING",
  "operatingHours": [
    {"type": "OPERATING", "startTime": "2026-08-12T11:30:00+02:00", "endTime": "2026-08-12T20:00:00+02:00"}
  ]
}
```

Status is derived from today's published window; the window itself ships under `operatingHours`, matching how dining is reported in `attractionsio`, `efteling`, `paultons` and `universal`. Like those, it is today only — the multi-day view stays in `getSchedules()`. No queue is attached, following the same reasoning as Europa-Park's `gastronomy_*` handling.

19 of the 50 restaurants get no row: counters and food trucks (`WEB Food Truck`, `Cool Post`, `Café Luminosity`, …) that Disney publishes no hours for. Reporting them closed all day would be a guess.

Measured across the day against the live feed:

| park time | OPERATING | CLOSED | REFURBISHMENT |
|---|---|---|---|
| 09:00 | 3 | 27 | 1 |
| 13:00 | 29 | 1 | 1 |
| 20:00 | 24 | 6 | 1 |
| 23:30 | 0 | 30 | 1 |

The window stays published outside opening hours, so a consumer polling at night still sees when the venue opens.

## REFURBISHMENT

This one status is passed through rather than folded into `CLOSED` by `mapStatus`, which is worth explaining because attractions do fold it.

The two sources mean different things. The wait feed reports `REFURBISHMENT` for anything not currently running — at 23:58 it covered 46 live attractions including *Big Thunder Mountain*, *Phantom Manor* and *Pirates of the Caribbean*. Folding that into `CLOSED` is correct.

The activity-schedule feed means it literally. Today it applies to exactly one venue, *Plaza Gardens Restaurant*, as a full-day `closed: true` row — and it is the only signal for that closure anywhere in the output, because `buildSchedules` skips `REFURBISHMENT` days and the venue therefore has **0 of 60 days** of published hours. Without this, a consumer sees `CLOSED` at lunchtime and cannot distinguish it from "shut right now".

Refurbishment is reported regardless of the time of day, and no `operatingHours` is attached — a `00:00–23:59` placeholder is not an opening window.

## Robustness

Schedule times arrive as plain GraphQL `String`s, so a malformed one is schema-legal. `"abc"` reached `constructDateTime`, which builds `2026-08-12Tabc` and throws — taking down the entire live build, not just one restaurant. Both the dining path and `isWithinWindow` (used by walkthrough attractions) now range-validate the time of day: shape and digit ranges both checked, `24:00` accepted as a midnight close, everything else read as CLOSED rather than thrown. A past-midnight close rolls into the next day, mirroring `buildSchedules`, and the window is half-open `[start, end)` to match `attractionsiov1`.

Fuzzed through the public path: missing / null / empty `schedules`, rows without times, empty and null time fields, non-numeric times, single-digit hours, out-of-range hours and minutes (`25:00`, `99:99`, `12:60`), unknown and null statuses, `closed` set against an OPERATING status, an end before its start, a full-day window, rows dated to another day, and two rows for today.

## Test plan

- [x] `npm test src/parks/dlp` — 26 tests across 2 files
- [x] `npm run dev -- disneylandparis --clear-cache` — 4/4, live rows 52 → 99
- [x] Verified against the live feed at several times of day, and at real time after every change

### DIFF
--- a/src/parks/dlp/__tests__/dining.test.ts
+++ b/src/parks/dlp/__tests__/dining.test.ts
@@ -1,0 +1,147 @@
+import {describe, it, expect, vi, beforeEach, afterEach} from 'vitest';
+import {DisneylandParis} from '../disneylandparis.js';
+
+/**
+ * Restaurants have no live feed, so their status is derived from today's
+ * published window. Anything Disney gives no hours for stays absent instead of
+ * being reported as closed all day.
+ */
+const TZ_OFFSET = '+02:00';
+
+function stubbedPark(hours: unknown, date: string): DisneylandParis {
+  const park = new DisneylandParis();
+
+  vi.spyOn(park as any, 'getPOIData').mockResolvedValue({
+    ThemePark: [{id: 'P1', name: 'Disneyland Park', type: 'ThemePark'}],
+    Restaurant: [{
+      id: 'P1RR01',
+      name: 'Casa de Coco',
+      type: 'Restaurant',
+      location: {id: 'P1'},
+      schedules: hours === undefined ? undefined : [{date, ...(hours as object)}],
+    }],
+  });
+  vi.spyOn(park as any, 'getWaitTimes').mockResolvedValue([]);
+  vi.spyOn(park as any, 'getPremierAccess').mockResolvedValue([]);
+  vi.spyOn(park as any, 'getVirtualQueueData').mockResolvedValue([]);
+  vi.spyOn(park as any, 'getScheduleForDate').mockResolvedValue([]);
+
+  return park;
+}
+
+/** Today in the park's timezone, as the park code computes it. */
+function parkToday(): string {
+  const [mm, dd, yy] = new Intl.DateTimeFormat('en-US', {
+    timeZone: 'Europe/Paris', month: '2-digit', day: '2-digit', year: 'numeric',
+  }).format(new Date()).split('/');
+  return `${yy}-${mm}-${dd}`;
+}
+
+async function statusAt(clock: string, hours: unknown): Promise<string | undefined> {
+  const date = parkToday();
+  vi.setSystemTime(new Date(`${date}T${clock}${TZ_OFFSET}`));
+  const live = await stubbedPark(hours, date).getLiveData();
+  return live.find((l) => l.id === 'P1RR01')?.status;
+}
+
+describe('DLP dining hours', () => {
+  beforeEach(() => {
+    vi.restoreAllMocks();
+    vi.useFakeTimers();
+  });
+  afterEach(() => vi.useRealTimers());
+
+  const operating = {startTime: '11:00:00', endTime: '21:30:00', status: 'OPERATING'};
+
+  it('reports OPERATING inside the published window', async () => {
+    expect(await statusAt('13:00:00', operating)).toBe('OPERATING');
+  });
+
+  it('reports CLOSED before opening', async () => {
+    expect(await statusAt('09:00:00', operating)).toBe('CLOSED');
+  });
+
+  it('reports CLOSED after closing', async () => {
+    expect(await statusAt('23:30:00', operating)).toBe('CLOSED');
+  });
+
+  it('reports REFURBISHMENT for a multi-day closure', async () => {
+    expect(await statusAt('13:00:00', {
+      startTime: '00:00:00', endTime: '23:59:00', status: 'REFURBISHMENT', closed: true,
+    })).toBe('REFURBISHMENT');
+  });
+
+  it('reports REFURBISHMENT regardless of the time of day', async () => {
+    const hours = {startTime: '00:00:00', endTime: '23:59:00', status: 'REFURBISHMENT', closed: true};
+    expect(await statusAt('03:00:00', hours)).toBe('REFURBISHMENT');
+    expect(await statusAt('23:30:00', hours)).toBe('REFURBISHMENT');
+  });
+
+  it('emits no row when no hours are published', async () => {
+    expect(await statusAt('13:00:00', undefined)).toBeUndefined();
+  });
+
+  it('emits no row when the window has no times', async () => {
+    expect(await statusAt('13:00:00', {status: 'OPERATING'})).toBeUndefined();
+  });
+
+  it('survives a malformed time instead of failing the whole build', async () => {
+    expect(await statusAt('13:00:00', {
+      startTime: 'abc', endTime: 'xyz', status: 'OPERATING',
+    })).toBe('CLOSED');
+  });
+
+  it('rejects a single-digit hour rather than truncating it', async () => {
+    // '9:30:00' sliced to HH:MM yields '9:30:', which is not a parseable time.
+    expect(await statusAt('13:00:00', {
+      startTime: '9:30:00', endTime: '20:00:00', status: 'OPERATING',
+    })).toBe('CLOSED');
+  });
+
+  it('ignores hours published for another day', async () => {
+    const date = parkToday();
+    vi.setSystemTime(new Date(`${date}T13:00:00${TZ_OFFSET}`));
+    const park = stubbedPark(operating, '2020-01-01');
+    const live = await park.getLiveData();
+    expect(live.find((l) => l.id === 'P1RR01')).toBeUndefined();
+  });
+
+  it('publishes today\'s window under operatingHours', async () => {
+    const date = parkToday();
+    vi.setSystemTime(new Date(`${date}T13:00:00${TZ_OFFSET}`));
+    const live = await stubbedPark(operating, date).getLiveData();
+
+    expect(live.find((l) => l.id === 'P1RR01')?.operatingHours).toEqual([{
+      type: 'OPERATING',
+      startTime: `${date}T11:00:00${TZ_OFFSET}`,
+      endTime: `${date}T21:30:00${TZ_OFFSET}`,
+    }]);
+  });
+
+  it('still publishes the window outside opening hours', async () => {
+    const date = parkToday();
+    vi.setSystemTime(new Date(`${date}T23:30:00${TZ_OFFSET}`));
+    const live = await stubbedPark(operating, date).getLiveData();
+    const entry = live.find((l) => l.id === 'P1RR01');
+
+    expect(entry?.status).toBe('CLOSED');
+    expect(entry?.operatingHours).toHaveLength(1);
+  });
+
+  it('omits operatingHours for a refurbishment', async () => {
+    const date = parkToday();
+    vi.setSystemTime(new Date(`${date}T13:00:00${TZ_OFFSET}`));
+    const live = await stubbedPark({
+      startTime: '00:00:00', endTime: '23:59:00', status: 'REFURBISHMENT', closed: true,
+    }, date).getLiveData();
+
+    expect(live.find((l) => l.id === 'P1RR01')?.operatingHours).toBeUndefined();
+  });
+
+  it('does not attach a queue to dining', async () => {
+    const date = parkToday();
+    vi.setSystemTime(new Date(`${date}T13:00:00${TZ_OFFSET}`));
+    const live = await stubbedPark(operating, date).getLiveData();
+    expect(live.find((l) => l.id === 'P1RR01')?.queue).toBeUndefined();
+  });
+});
--- a/src/parks/dlp/__tests__/dining.test.ts
+++ b/src/parks/dlp/__tests__/dining.test.ts
@@ -1,33 +1,13 @@
 import {describe, it, expect, vi, beforeEach, afterEach} from 'vitest';
 import {DisneylandParis} from '../disneylandparis.js';
+import {constructDateTime} from '../../../datetime.js';
 
 /**
  * Restaurants have no live feed, so their status is derived from today's
- * published window. Anything Disney gives no hours for stays absent instead of
- * being reported as closed all day.
+ * window in the activity-schedule feed (the POI blob's schedules only carry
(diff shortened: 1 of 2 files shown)


cubehouse (changes requested):
The dining feature is worth having and the REFURBISHMENT reasoning in the description is right. I verified that part: in-park restaurants produce exactly two statuses across a 60-day sweep, `OPERATING` (1599 rows) and `REFURBISHMENT` (243), and `buildSchedules` drops the REFURBISHMENT days at `:1131`, so the live status genuinely is the only signal. Passing it through rather than folding it into `CLOSED` is the better call.

The headline also checks out, measured against the live feed on the same cache and the same minute: live rows 54 to 83, RESTAURANT rows 0 to 31, all carrying `operatingHours` except the single REFURBISHMENT row, none carrying a queue, and no other entity type affected.

Three things need fixing.

## Blocking

### 1. Every restaurant row disappears for up to 12 hours after local midnight

`disneylandparis.ts:1075` reads `poi.schedules`, which comes from `getPOIData()` at `:393` under `@cache({ttlSeconds: 43200})`. The POI blob's `schedules` array only ever contains **one date**, the day it was fetched. So once the park clock rolls past midnight there is no row for the new date and every restaurant is skipped.

Measured at 00:03 Europe/Paris against a POI cache filled at 23:53:

```
todayStr 2026-08-13
POI (12h-cached) in-park restaurants with a row for todayStr:  0 of 56
activity-schedule feed for todayStr, in-park restaurants:     33
```

Live rows dropped 83 to 68, all 31 restaurant rows gone. A forced refetch at 00:04 returned `{'2026-08-13': 375}`, so upstream rolls over at midnight correctly and the entire gap is our own cache TTL. Depending on where the 12h boundary falls, dining goes dark from local midnight until as much as 12h later, straight through park open and past most restaurants' 11:00 opening.

The fix is already in the same method. `getScheduleForDate(todayStr)` is fetched at `:917` for showtimes and carries every in-park restaurant with correct `startTime`, `endTime`, `status` and `closed` for the actual current date (33 entities at 00:03, including the REFURBISHMENT rows). Sourcing dining from that payload rather than `poi.schedules` removes the dependency entirely.

Note the walkthrough derivation at `:1032` has the same dependency and reports CLOSED after midnight for the same reason. That one predates this PR, so it is not yours to fix here, but this change makes the flaw load-bearing for 31 more entities.

### 2. `TIME_OF_DAY` accepts values that then throw

`disneylandparis.ts:56`, used at `:1023` and `:1087`. `/^\d{2}:\d{2}/` accepts `25:00`, `99:99` and `12:60`, and `constructDateTime` throws `RangeError: Invalid time value` on all three. Driven through the public `getLiveData()` with a stubbed POI, `25:00:00` and `99:99:99` reject the whole promise, on the dining path and the walkthrough path. That is the exact failure the guard's own comment says it prevents.

Adding three `it.each` cases to `dining.test.ts` fails immediately with `RangeError`, which is the shortest proof.

#296 has a stricter pattern, `/^([01]\d|2[0-3]):[0-5]\d(:[0-5]\d)?$/`. One caveat before adopting it wholesale: it rejects `24:00:00`, which the current loose regex handles correctly and which is a plausible midnight-close encoding. Either accept `24:00` explicitly or take that regression knowingly.

Credit where it is due: the guard does close a real pre-existing hole. An `undefined` startTime previously reached `open.slice()` and threw `TypeError`, and `TIME_OF_DAY.test(undefined)` now returns false.

### 3. The `as any` casts should go

`disneylandparis.ts:1084`, `:1088`, `:1099`. I removed all three and `tsc --noEmit` exits 0, so they are not needed. `LiveStatusTypeEnum` includes `REFURBISHMENT`, and seaworld, efteling and shdr all assign it uncast.

Worth insisting on rather than treating as style: `Destination.getLiveData()` sanitises only `waitTime`. Nothing validates `status`, so the compiler is the only line of defence and the cast removes it.

## Test findings

The stated test plan says the schedule payloads were fuzzed. `dining.test.ts:88-99` is the entire malformed-input coverage, and it is `'abc'`, `'xyz'` and `'9:30:00'`, all string-shape failures the loose regex already catches. There is no out-of-range case anywhere in the file, which is why finding 2 shipped.

Ran the suite against `main`'s source: **9 failed, 5 passed**. Two of the five passers are actively misleading:

- `:131` `omits operatingHours for a refurbishment` and `:141` `does not attach a queue to dining` both use `live.find(...)?.operatingHours` / `?.queue`. When the row is absent the optional chain yields `undefined` and the assertion passes, so both pass with the feature deleted entirely. Assert the row exists first.

Two mutations that should fail and do not:

| Mutation at `:1095-1098` | Result |
|---|---|
| `hours.closed !== true` to `true` | 14/14 pass |
| delete `&& this.mapStatus(hours.status) === 'OPERATING'` | 14/14 pass |

`closed` is only exercised through the REFURBISHMENT fixtures, which `return` at `:1083` before reaching the expression. No test supplies a non-OPERATING non-REFURBISHMENT status, or a `closed: true` OPERATING row.

Separately, `dining.test.ts:9` hardcodes `TZ_OFFSET = '+02:00'` while the code derives the offset from the real current date. Pinning the clock to December fails `publishes today's window under operatingHours` and nothing else, so **the suite breaks on 2026-10-25 and stays broken until 2027-03-29**. Derive the offset from the produced value or assert against `constructDateTime` output.

## Smaller notes

- **Unknown status maps to OPERATING.** `mapStatus`'s default branch at `:678` is borrowed from the wait feed. Any new token reports OPERATING inside the window, and `DOWN` maps to `'DOWN'`, fails the `=== 'OPERATING'` test and publishes as CLOSED, losing the signal. An explicit allowlist fits this call site better.
- **`hours.closed === true` is redundant today**: 0 rows in 60 days where `status === 'OPERATING'` disagrees with `closed !== true`. Harmless, but see the untested mutation above.
- **Midnight crossing**: 377 crossing rows in the 60-day feed, all hotel and Disney Village, which `filterPOIEntities` drops. Zero for in-park restaurants, so this is latent only. Worth a line of code though, since `buildSchedules` explicitly handles the shape at `:1136-1143` and the new block does not. An in-park `endTime: '01:00:00'` on a party night would read CLOSED all day and publish an inverted slot.
- **Two rows for one date**: `.find()` at `:1075` takes the first. Zero occurrences in 60 days, but a lunch/dinner split would collapse to the first window and read CLOSED through service.
- **Dead branch**: `:1079` `if (!ld) continue;` cannot fire, since `poi` comes from `emittablePois`, which is the set backing `validEntityIds`.
- **Inclusive end**: `nowMs <= endTime` at `:1098` vs the half-open `[start, end)` in `attractionsiov1.ts:303-309`. A one-minute inconsistency across parks.
- **Policy inconsistency**: the walkthrough block defers to upstream feeds at `:1066-1067` ("those feeds are authoritative") while the dining block unconditionally overwrites `status`. Unreachable today, but the two blocks should state the same rule.
- DLP now reports refurbishment two ways: attractions fold it into CLOSED at `:676`, restaurants pass it through. Same park, same upstream token, two meanings. Passing it through is the better choice, so the attraction mapping should probably follow, or the description should say why not.

## Not yours to fix

The branch is behind `main` and needs #292 before the numbers are re-measured. While reviewing this I also found a regression that came in with #292, not with your PR: `parkCloseStr` at `:1015-1017` takes the max `endTime` across queue-bearing attractions, and #292 made the railroad stations eligible to contribute, so on days a station publishes a late OPERATING window every walkthrough attraction reads OPERATING past park close. I will pick that up separately.

BeyondVertical:
All three blocking findings are fixed in a2c00fba, and `main` (with #292) is merged in.

**1. The midnight gap** — dining now reads from the `getScheduleForDate(todayStr)` payload the showtimes path already fetches, exactly as you suggested. The POI-blob dependency is gone entirely; a regression test pins it (POI carries only yesterday's row, dining still publishes). The walkthrough derivation at the park-hours fallback keeps its POI dependency as you noted — not touched here.

**2. `TIME_OF_DAY`** — now range-validating, with `24:00` explicitly accepted rather than regressed (`24:00(?::00)?` alternates with the strict `([01]\d|2[0-3]):[0-5]\d` form). Your three `it.each` cases (`25:00`, `99:99`, `12:60`) are in the suite and read CLOSED through the public path instead of rejecting the promise.

**3. The casts** — all three gone; `tsc --noEmit` clean.

Test findings, all taken:

- The timezone offset is derived from `constructDateTime` output instead of hardcoded, so the suite survives 2026-10-25.
- The refurbishment and queue tests assert the row exists before the optional chain.
- Both surviving mutations now go red: `closed: true` against an OPERATING row and an unrecognised status token each have a test.
- 20 of 23 dining tests fail against the pre-fix source.

From the smaller notes:

- Unknown status → explicit allowlist: only a literal `OPERATING` opens the row, so a new token no longer borrows the wait feed's OPERATING default. `DOWN` reads CLOSED — with no other signal published for dining I kept that rather than inventing a pass-through for a token the feed has never emitted here.
- Midnight crossing rolls the close into the next day, mirroring `buildSchedules`' handling, with a test for a `19:00 → 01:00` window.
- Inclusive end fixed to half-open `[start, end)`, matching `attractionsiov1`; boundary minutes are under test.
- The overwrite-vs-defer inconsistency is now stated as a rule in the comment: the dining feed is authoritative for a restaurant's status.
- `hours.closed !== true` stays, and is now load-bearing in a test rather than only through the refurbishment fixtures.
- The `if (!ld) continue;` after `getOrCreate` stays for type narrowing — `getOrCreate` is typed `LiveData | undefined`, so removing it trades dead code for a non-null assertion.
- On attractions folding REFURBISHMENT into CLOSED while dining passes it through: kept deliberately, and the description now spells out why — the wait feed parks the whole estate on REFURBISHMENT overnight (46 rows at 23:58), so for attractions the token is an idle state, while the schedule feed's is a real multi-day closure.

Description updated for the new data source, the fuzz list (out-of-range cases included now) and the test counts.

Verified live at 13:33 Paris: 31 restaurant rows of 50 entities, 29 OPERATING / 1 REFURBISHMENT / 1 CLOSED, `operatingHours` on all but the refurbishment row, zero queues.

cubehouse (approved):
All three blocking findings are addressed. Approving.

Measured against the live feed on 2026-08-14 between 18:28 and 18:40 Paris, both parks open on published hours 09:30 to 22:00.

## The three blockers

**1. Post-midnight blackout, fixed.** Dining now reads `todaySchedules` from the activity-schedule payload the showtimes path already fetches, hoisted at `:926` and reused at `:1088`. The POI blob dependency is gone, so the 12h cache can no longer strand every restaurant after local midnight.

**2. `TIME_OF_DAY` range validation, fixed.** The new pattern rejects `25:00`, `99:99` and `12:60` while keeping `24:00` as a midnight close. I checked the `24:00` half rather than taking it on trust: `new Date('2026-08-14T24:00:00Z')` gives `2026-08-15T00:00:00.000Z`, and `constructDateTime`'s wall-clock convergence loop lands on Paris midnight from there. It works.

**3. The three `as any` casts are gone.** `status` is compiler-checked again, which matters because the base class sanitises only `waitTime`.

## Live verification of the dining path

| | today |
|---|---|
| restaurant rows in the schedule feed | 30 `OPERATING`, 1 `REFURBISHMENT` |
| restaurants with two rows for one date | 0 |
| non-`OPERATING` rows carrying times | 0 |
| emittable restaurants with no row, correctly skipped | 19 of 50 |

That matches the 31 RESTAURANT live rows from the earlier measurement exactly.

## One latent note, not blocking

`:1093` `.find((s) => s.date === todayStr)` takes the first row of the date whatever its status, and `:1128` then writes `operatingHours: [{type: 'OPERATING', ...}]` even on a row that was `CLOSED`. Nothing reaches it today (0 such rows above), but a lunch/dinner split or a `CLOSED` row carrying times would publish a closure window labelled as opening hours. Narrowing the find closes it:

```ts
.find((s) => s.date === todayStr && (s.status === 'OPERATING' || s.status === 'REFURBISHMENT'))
```

## Merge notes

- **`TIME_OF_DAY` is defined differently here and in #296**, so those two conflict loudly on a duplicate `const`. Keep this one. It is the superset, and `buildSchedules` passes the full `HH:MM:SS` string to `constructDateTime`, which handles `24:00:00` correctly per the check above.
- **#298 adds a third `getScheduleForDate(todayStr)` call site.** Cached at 24h so no extra request, but collapse them onto the `todaySchedules` this PR introduces when the two merge.
- #298 also fixes the walkthrough half of the same POI staleness you flagged here, and leaves `deriveWalkthroughStatus`'s own-hours lookup on `poi.schedules`. `todaySchedules` is the fix for that too, which is another reason to land these two together.
- Half-open `[start, end)` here at `:1126` versus #298's inclusive `nowMs <= c`. Worth settling on one when they merge.

PR #295, opened 2026-08-11
feat(dlp): publish attraction tags from the POI feed

BeyondVertical:
## Summary

`getEntities()` returned `tags: []` for all 129 entities. The tag code read `poi.minimumHeight` and `poi.height[].value`, but the POI query requested neither — and `minimumHeight` is not a field on the GraphQL `Attraction` type at all, so it could never have been requested.

`activities()` returns the `Activity` interface, so concrete-type fields need an inline fragment. Behind `... on Attraction` sit the values these tags were always meant to carry.

| | before | after |
|---|---|---|
| Entities carrying ≥1 tag | 0 | **23** of 129 |
| Tags emitted | 0 | **54** |
| DLP tests | 2 | **67** |

Nothing else moves: 129 entities in, 129 out, same counts per type, and the schedule path is untouched. No new HTTP request either — the fields ride along on the POI query that already runs. A second commit follows on from the same fragment and touches the live-data path; it is described in a comment below.

| tag | attractions | source |
|---|---|---|
| `PAID_RETURN_TIME` | 19 | `interests` → `disney-premier-access-one` |
| `MINIMUM_HEIGHT` | 10 | `height[].id` |
| `SINGLE_RIDER` | 9 | `singleRider` |
| `ONRIDE_PHOTO` | 7 | `interests` → `PhotoPass` |
| `UNSUITABLE_PREGNANT` | 5 | `physicalConsiderations` → `expectantMothersMayNotRide` |
| `MAY_GET_WET` | 4 | `interests` → `guestMayGetSplashed` |

No new tag type is introduced; all six already exist in `src/tags/tagTypes.ts`.

```json
{
  "id": "P1RA00",
  "name": "Big Thunder Mountain",
  "entityType": "ATTRACTION",
  "tags": [
    {"tag": "MINIMUM_HEIGHT", "tagName": "Minimum Height", "value": {"height": 102, "unit": "cm"}},
    {"tag": "UNSUITABLE_PREGNANT", "tagName": "Unsuitable for Pregnant People"},
    {"tag": "PAID_RETURN_TIME", "tagName": "Paid Return Time"},
    {"tag": "ONRIDE_PHOTO", "tagName": "On-Ride Photo"}
  ]
}
```

## Height is parsed from the facet id, not its value

Each `height` entry has an `id` and a display `value`. The value is localised — the same restriction reads `1.20 m` under `en-gb` and `1,20 m` under `fr-fr` — while the ids (`anyHeight`, `81cm`, `1_02m`, `1_07m`, `1_20m`, `1_40m`) are identical across every market checked. Since `DLP_LANGUAGE` is configurable, a value-based parser would report `1,20 m` as 100 cm under a French market. This matches how `shdr` reads its height facet.

The ten restrictions published today: 81 cm (Autopia, Toy Soldiers Parachute Drop), 1.02 m (Big Thunder Mountain, Star Tours, Tower of Terror), 1.07 m (Crush's Coaster), 1.20 m (Flight Force, RC Racer, Hyperspace Mountain), 1.40 m (Indiana Jones).

## Premier Access

`interests` carries three Premier Access facets. `disney-premier-access-one` and `orion` name the same 19 attractions; `disney-premier-access-ultimate` names 18 of those. Those 19 are exactly the 19 `attractionId`s the live Premier Access endpoint returns — no difference in either direction — so the tag agrees by construction with the `queue.PAID_RETURN_TIME` the live path already emits.

## PhotoPass → ONRIDE_PHOTO

Worth flagging, because the facet names the service rather than the camera. Ten POI entries carry `PhotoPass`. Seven are emitted, and all seven are rides with an on-ride camera: Flight Force, Big Thunder Mountain, Buzz Lightyear Laser Blast, Frozen Ever After, Pirates of the Caribbean, Hyperspace Mountain, Tower of Terror. The other three are walk-up photo locations — *Meet Mickey Mouse*, *Princess Pavilion*, *Welcome to Starport* — all `-OLD` entries flagged `Hide from Web List + Mobile App`, so the visibility filter already drops them. The mapping is therefore exact over the published set, but it would over-tag if Disney ever surfaces a visible walk-up PhotoPass location. Happy to drop this one line if you'd rather not carry that.

## What the feed offers and this deliberately leaves alone

- `fastPass` and `virtualQueue` are real Boolean fields on `Attraction`, but `false` on all 62 attractions — `fastPass` included, on the same 19 rides whose `interests` and live feed both say Premier Access is sold. Neither flag tracks reality, so neither is used.
- `physicalConsiderations` also publishes `mustBeInGoodHealth` (3), `typesOfLimbImpairement` and `typesOfLimbImpairementText` (1 each). No existing tag type fits, and inventing one is out of scope here.
- `subLocation` names the land (Fantasyland, Worlds of Pixar, …) for 61 of 62 attractions. There is no `LAND` entity type and no location tag that fits.
- `coordinates` only ever carries `Guest Entrance` for attractions, which is already `entity.location`. Per `TAG_DEVELOPMENT_GUIDE.md`, no `LOCATION` tag.
- `minimumHeight`, `maximumHeight`, `thrills`, `photoPass`, `onRidePhoto`, `mayGetWet`, `childSwap` and `description` were each probed against the schema and every one comes back `FieldUndefined` on `Attraction`.

## Removed

`parseHeightCm` as a method, the `minimumHeight` type member, and the `value`/`iconFont` members of `height[]` — none of them reachable, and nothing else in the repo reads them. The parser returns as a module-level helper next to the type it parses, alongside `parseDLPWait` and `parseDLPDate`.

`getPOIData`'s `cacheVersion` goes 2 → 3 so warm caches don't serve the pre-fragment shape after deploy. The HTTP layer keys on the request body, so that layer invalidates itself.

One thing to watch on merge: #293 bumps the same `cacheVersion` to the same 3, and both branches add a module constant right after `SHOW_SUBTYPES` and restructure the same `DLPPOIEntity` block. The constant and the type block do conflict, so those can't slip through — but **`cacheVersion` does not conflict**, because both sides make the identical edit, so git resolves it silently to 3. Whichever lands second needs taking to `cacheVersion: 4` by hand, since by then the shape will have changed twice. Verified by merging the branches locally.

## Robustness

`buildTags` runs inside `buildEntityList`, so one bad record would take down the whole entity list rather than a single attraction, and a bad facet id would put nonsense in the output rather than nothing.

The malformed shapes are driven through the public `getEntities()` path, none of which throws or produces a tag: the three facet lists as `null`, as a string, as a number, as a bare object, containing `null` entries, containing entries with no `id`, and with non-string ids; `singleRider` as `'yes'`, `1`, `{}`; and `Attraction` itself as `null`, a string, `[null]`, `[{}]`, `[42]`. Repeating a facet 500 times still yields one tag.

`parseHeightCm` accepts only `anyHeight`-style ids it recognises. Rejected: the empty string, whitespace, `0cm`, `0_00m`, `-5cm`, `1e3cm`, `1_2_3m`, a bare `m`, an unterminated `1_20`, an imperial `36in`, and the case- and space-variants `81CM` and `" 81cm "`. Two guards do real work:

- `cm > 0`, because `isHeightValue` accepts `height >= 0` — a `0cm` facet would otherwise ship a valid-but-meaningless `{height: 0, unit: 'cm'}`.
- `cm <= 300`, matching `oceanpark`'s `HEIGHT_NO_LIMIT_CM`, so a mistyped `999999cm` is dropped instead of published. `300cm` itself still passes.

One case does throw, and it predates this change: `flattenPOI` calls `Object.entries` on its argument, so a literally `null` POI response fails. It is unreachable — `getPOIData` returns `data?.data || {}` — and out of scope here.

## Test plan

- [x] `npm test src/parks/dlp` — 67 tests across 3 files
- [x] `npm run dev -- disneylandparis --clear-cache` — 4/4, 129 entities, unchanged counts per type
- [x] Full suite: 1744 passing; the two failures in `cache.test.ts` / `cachePragmas.test.ts` are pre-existing on `main` (SQLite WAL pragmas on Windows)
- [x] Every emitted tag passes `TagBuilder.validateAll()`; no non-`ATTRACTION` entity carries one

### DIFF
--- a/src/parks/dlp/__tests__/singleRider.test.ts
+++ b/src/parks/dlp/__tests__/singleRider.test.ts
@@ -3,9 +3,10 @@ import {DisneylandParis} from '../disneylandparis.js';
 import {CacheLib} from '../../../cache.js';
 
 /**
- * Single-rider capability comes from the POI `singleRider` facet. The wait
- * feed reports the same flag, but only answers while the park is awake, so
- * the facet is what keeps a ride's queue present overnight.
+ * Single-rider capability comes from the POI `singleRider` facet, OR'd with
+ * wait-feed sightings from the last 48h. The facet keeps a ride's queue
+ * present overnight; the sighting bridges the POI cache lagging a fresh
+ * addition, and lapses so a retired queue disappears.
  */
 function stubbedPark(
   attractions: Array<Record<string, unknown>>,
@@ -91,6 +92,34 @@ describe('DLP single rider', () => {
       [feedRow({singleRider: {isAvailable: false}})],
     ));
     expect(queue.SINGLE_RIDER).toEqual({waitTime: null});
+  });
+
+  it('bridges a fresh addition the POI cache does not carry yet', async () => {
+    // The feed advertised single rider this morning; the facet lags behind.
+    await stubbedPark(
+      [RIDE],
+      [feedRow({singleRider: {isAvailable: true, singleRiderWaitMinutes: 10}})],
+    ).getLiveData();
+
+    const queue = await queueOf(stubbedPark([RIDE]));
+    expect(queue.SINGLE_RIDER).toEqual({waitTime: null});
+  });
+
+  it('lets a retired single-rider queue disappear once the sighting lapses', async () => {
+    vi.useFakeTimers();
+    try {
+      await stubbedPark(
+        [{...RIDE, singleRider: true}],
+        [feedRow({singleRider: {isAvailable: true, singleRiderWaitMinutes: 10}})],
+      ).getLiveData();
+
+      // Two days on: no facet, feed asleep, last sighting lapsed.
+      vi.setSystemTime(Date.now() + 49 * 60 * 60 * 1000);
+      const queue = await queueOf(stubbedPark([RIDE]));
+      expect(queue.SINGLE_RIDER).toBeUndefined();
+    } finally {
+      vi.useRealTimers();
+    }
   });
 
   it.each([
--- a/src/parks/dlp/__tests__/singleRider.test.ts
+++ b/src/parks/dlp/__tests__/singleRider.test.ts
@@ -1,0 +1,107 @@
+import {describe, it, expect, vi, beforeEach} from 'vitest';
+import {DisneylandParis} from '../disneylandparis.js';
+import {CacheLib} from '../../../cache.js';
+
+/**
+ * Single-rider capability comes from the POI `singleRider` facet. The wait
+ * feed reports the same flag, but only answers while the park is awake, so
+ * the facet is what keeps a ride's queue present overnight.
+ */
+function stubbedPark(
+  attractions: Array<Record<string, unknown>>,
+  waitTimes: unknown[] = [],
+): DisneylandParis {
+  const park = new DisneylandParis();
+
+  vi.spyOn(park as any, 'getPOIData').mockResolvedValue({
+    ThemePark: [{id: 'P1', name: 'Disneyland Park', type: 'ThemePark'}],
+    Attraction: attractions.map((a) => ({
+      type: 'Attraction',
+      location: {id: 'P1'},
+      ...a,
+    })),
+  });
+  vi.spyOn(park as any, 'getWaitTimes').mockResolvedValue(waitTimes);
+  vi.spyOn(park as any, 'getPremierAccess').mockResolvedValue([]);
+  vi.spyOn(park as any, 'getVirtualQueueData').mockResolvedValue([]);
+  vi.spyOn(park as any, 'getScheduleForDate').mockResolvedValue([]);
+
+  return park;
+}
+
+const RIDE = {id: 'P1RA00', name: 'Big Thunder Mountain'};
+
+/** A wait-feed row, which is what marks a ride as queue-bearing. */
+const feedRow = (over: Record<string, unknown> = {}) => ({
+  entityId: 'P1RA00',
+  type: 'Attraction',
+  status: 'OPERATING',
+  postedWaitMinutes: 30,
+  ...over,
+});
+
+async function queueOf(park: DisneylandParis): Promise<any> {
+  const live = await park.getLiveData();
+  const row = live.find((l) => l.id === 'P1RA00');
+  expect(row).toBeDefined();
+  return (row as any)?.queue;
+}
+
+describe('DLP single rider', () => {
+  beforeEach(() => {
+    vi.restoreAllMocks();
+    CacheLib.clearByClassName('DisneylandParis');
+  });
+
+  it('emits the queue from the POI facet while the feed is asleep', async () => {
+    // The ride is known to be queue-bearing from an earlier live feed, but
+    // right now the feed returns nothing at all.
+    await stubbedPark([{...RIDE, singleRider: true}], [feedRow()]).getLiveData();
+
+    const queue = await queueOf(stubbedPark([{...RIDE, singleRider: true}]));
+    expect(queue.SINGLE_RIDER).toEqual({waitTime: null});
+  });
+
+  it('emits no queue for a ride the facet does not flag', async () => {
+    await stubbedPark([RIDE], [feedRow()]).getLiveData();
+
+    const queue = await queueOf(stubbedPark([RIDE]));
+    expect(queue.SINGLE_RIDER).toBeUndefined();
+  });
+
+  it('lets the live feed set the wait time', async () => {
+    const queue = await queueOf(stubbedPark(
+      [{...RIDE, singleRider: true}],
+      [feedRow({singleRider: {isAvailable: true, singleRiderWaitMinutes: 15}})],
+    ));
+    expect(queue.SINGLE_RIDER).toEqual({waitTime: 15});
+  });
+
+  it('nulls the wait time when the ride is closed', async () => {
+    const queue = await queueOf(stubbedPark(
+      [{...RIDE, singleRider: true}],
+      [feedRow({status: 'CLOSED', singleRider: {isAvailable: true, singleRiderWaitMinutes: 15}})],
+    ));
+    expect(queue.SINGLE_RIDER).toEqual({waitTime: null});
+  });
+
+  it('keeps the queue when the live feed reports the flag false', async () => {
+    const queue = await queueOf(stubbedPark(
+      [{...RIDE, singleRider: true}],
+      [feedRow({singleRider: {isAvailable: false}})],
+    ));
+    expect(queue.SINGLE_RIDER).toEqual({waitTime: null});
+  });
+
+  it.each([
+    ['a string', 'true'],
+    ['a number', 1],
+    ['null', null],
+    ['absent', undefined],
+  ])('emits no queue when the facet is %s', async (_label, singleRider) => {
+    await stubbedPark([{...RIDE, singleRider}], [feedRow()]).getLiveData();
+
+    const queue = await queueOf(stubbedPark([{...RIDE, singleRider}]));
+    expect(queue.SINGLE_RIDER).toBeUndefined();
+  });
+});
--- a/src/parks/dlp/__tests__/tags.test.ts
+++ b/src/parks/dlp/__tests__/tags.test.ts
@@ -1,0 +1,202 @@
+import {describe, it, expect, vi, beforeEach} from 'vitest';
+import {DisneylandParis} from '../disneylandparis.js';
+
+/**
+ * Every field the tags are built from lives on the GraphQL `Attraction` type,
+ * reached through an inline fragment. Anything else in the POI feed therefore
+ * carries no tags at all.
+ */
+function stubbedPark(attraction: Record<string, unknown>): DisneylandParis {
+  const park = new DisneylandParis();
+
+  vi.spyOn(park as any, 'getPOIData').mockResolvedValue({
+    ThemePark: [{id: 'P1', name: 'Disneyland Park', type: 'ThemePark'}],
+    Attraction: [{
+      id: 'P1RA00',
+      name: 'Big Thunder Mountain',
+      type: 'Attraction',
+      location: {id: 'P1'},
+      ...attraction,
+    }],
+    Entertainment: [{
+      id: 'P1GS00',
+      name: 'Disney Illuminations',
+      type: 'Entertainment',
+      subType: 'Fireworks',
+      location: {id: 'P1'},
+    }],
+    Restaurant: [{
+      id: 'P1AR06',
+      name: 'Casa de Coco',
+      type: 'Restaurant',
+      location: {id: 'P1'},
+    }],
+  });
+
+  return park;
+}
+
+async function tagsOf(attraction: Record<string, unknown>): Promise<string[]> {
+  const entities = await stubbedPark(attraction).getEntities();
+  const target = entities.find((e) => e.id === 'P1RA00');
+
+  // Without this, every `toEqual([])` below would also pass on an attraction
+  // that never made it into the output at all.
+  expect(target).toBeDefined();
+
+  return (target?.tags ?? []).map((t) => t.tag);
+}
+
+describe('DLP attraction tags', () => {
+  beforeEach(() => vi.restoreAllMocks());
+
+  it('emits no tags for an attraction without any of the fields', async () => {
+    expect(await tagsOf({})).toEqual([]);
+  });
+
+  it('emits no tags for shows or restaurants', async () => {
+    const entities = await stubbedPark({}).getEntities();
+    expect(entities.find((e) => e.id === 'P1GS00')?.tags).toEqual([]);
+    expect(entities.find((e) => e.id === 'P1AR06')?.tags).toEqual([]);
+  });
+
+  // === Minimum height ===
+
+  it.each([
+    ['81cm', 81],
+    ['1_02m', 102],
+    ['1_07m', 107],
+    ['1_20m', 120],
+    ['1_40m', 140],
+  ])('maps the %s facet to %i cm', async (facetId, expected) => {
+    const entities = await stubbedPark({height: [{id: facetId}]}).getEntities();
+    const tag = entities.find((e) => e.id === 'P1RA00')?.tags?.[0];
+
+    expect(tag?.tag).toBe('MINIMUM_HEIGHT');
+    expect(tag?.value).toEqual({height: expected, unit: 'cm'});
+  });
+
+  it('ignores the localised value string in favour of the facet id', async () => {
+    // The same facet reads "1,20 m" under a French market.
+    const entities = await stubbedPark({
+      height: [{id: '1_20m', value: '1,20 m'}],
+    }).getEntities();
+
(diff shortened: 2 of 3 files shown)


BeyondVertical:
Pushed a second commit — `fix(dlp): let a retired single-rider queue disappear`.

`singleRider` is the one facet in the table above that the live path was already deriving for itself. `buildLiveData` keeps a 30-day `dlp:singleRiderCapable` cache of every id the wait feed has reported as available, so the queue can be re-emitted with `waitTime: null` overnight. The comment justifying it says single-rider eligibility "isn't on POI" — which the inline fragment in this PR makes untrue, so I wanted to check whether the cache still earns its place.

The two sources were measured against each other across this morning, during Extra Magic Time and again with the park fully open:

| sample | wait-feed state | POI facet | wait feed | facet-only | feed-only |
|---|---|---|---|---|---|
| 08:37 | 16 OPERATING / 51 REFURBISHMENT | 9 | 10 | — | `P2YA03` |
| 08:45 | 16 / 51 | 9 | 10 | — | `P2YA03` |
| 09:34 | 46 / 21 | 9 | 10 | — | `P2YA03` |

Between the second and third sample thirty attractions opened and the single-rider waits became live, scattered values (0, 5, 5, 5, 5, 10, 10, 20, 45). The set itself never moved.

`P2YA03` is the only difference and it is unobservable: the id exists in none of the thirteen POI types, so `getOrCreate`'s emittable-set gate drops it. Eighteen further wait-feed ids are in the same position.

The flag is static rather than operational — Indiana Jones, RC Racer and Toy Soldiers each reported `isAvailable: true` while sitting on `REFURBISHMENT` ahead of their opening time. So it is the same quantity as the facet, not a live measurement of it.

What the cache does wrong is that it can never forget. `seenSR` seeds from its own cached value, only ever adds, and rewrites a fresh 30-day TTL on every run, so a single-rider queue Disney retires would keep being published for as long as the process runs:

| scenario | before | after |
|---|---|---|
| cold cache, feed asleep | 0 | 0 |
| live feed | 9 | 9 |
| warm queue-bearing history, feed asleep | 9 | 9 |
| cache seeded with two ids that carry no single-rider queue | **11** | **9** |

The first row is worth stating plainly: this is **not** a cold-start fix. The baseline emission is also gated on `queueBearingIds`, which is empty on a cold start too, so both versions emit nothing. The only behavioural difference is the last row.

No `cacheVersion` bump for this commit — the first one already put the field in the query, so the cached shape is unchanged. Net −12 lines in `disneylandparis.ts`.

Verification:

- 56 DLP tests across 3 files, `tsc --noEmit` clean, full suite 1744 passing (the two `cache.test.ts` / `cachePragmas.test.ts` failures are pre-existing on `main` — SQLite WAL pragmas on Windows)
- `npm run dev -- disneylandparis --clear-cache` — 4/4, entity and live-row counts unchanged
- Against the live feed with the park open, the `SINGLE_RIDER` tag and `queue.SINGLE_RIDER` name the same nine rides, and `PAID_RETURN_TIME` the same nineteen — no difference in either direction
- The two tests that encode the change fail against the previous implementation; the other seven pass either way, as guardrails should

Happy to split this into its own PR once #295 lands if you'd rather keep them separate — it only sits here because this branch is what puts `singleRider` in the query and the type.

cubehouse (changes requested):
The tag work is solid and the test suite is real rather than decorative: run against `main`'s source, 20 of 54 fail, so most assertions are load-bearing. Live output checks out too, verified against the feed: 129 entities, 23 tagged, all 23 `ATTRACTION`, every tag type valid per `src/tags/tagTypes.ts`, every `MINIMUM_HEIGHT` in the 81 to 140cm band. `PAID_RETURN_TIME` agreement is exact, 19 tagged ids vs 19 `queue.PAID_RETURN_TIME` ids, nothing in either direction only. Same for `SINGLE_RIDER`, 9 vs 9.

Your call on reading height from the facet id rather than the localised `value` is right, and the note about `disney-premier-access-ultimate` being the wrong marker is right too.

Three things to fix.

## Blocking

### 1. `parseHeightCm` publishes a nonsense 1cm restriction on mixed-unit ids

`disneylandparis.ts:116-123`. Every input the description claims to reject **is** rejected, verified through the public `getEntities()` path along with 25 shapes of my own (`+81cm`, `81 cm`, `81cm\n`, `1_20M`, `_20m`, `1_m`, `0m`, `00cm`, `1_20mm`, and a 20-digit value). The gap is the `cm > 0` bound at `:122`, because `Math.round` runs first:

| input | result |
|---|---|
| `1_20cm` | **`MINIMUM_HEIGHT(1cm)`** |
| `0_81cm` | **`MINIMUM_HEIGHT(1cm)`** |
| `1cm` | `MINIMUM_HEIGHT(1cm)` |
| `2_50m` | `MINIMUM_HEIGHT(250cm)` |
| `3m` | `MINIMUM_HEIGHT(300cm)` |

`1_20cm` is the interesting one, since an underscore-decimal paired with the `cm` unit is exactly the mixed shape the rejection list was reaching for, and it ships a 1cm minimum height to the wiki. `MAX_HEIGHT_CM = 300` is a defensible bad-data ceiling but it is twice any real minimum. A plausible-rider band, say 40 to 200, closes both ends.

For reference, the live feed carries exactly 6 distinct height facet ids: `anyHeight` (51), `1_02m` (3), `1_20m` (3), `81cm` (2), `1_07m` (1), `1_40m` (1). All handled, and zero attractions carry two height facets, so the first-match behaviour does not bite today. Worth knowing that `interests` contains a stray `1_20m` id, so facet namespaces do overlap across lists; harmless while height is read only from `poi.height`.

### 2. The second commit has no test for the bug it is named after

`3efd1da8` is titled "let a retired single-rider queue disappear", but only 2 of the 10 `singleRider.test.ts` tests fail against pre-PR source, and neither is the retirement case. `singleRider.test.ts:66-71` warms the cache with `feedRow()`, which carries no `singleRider` payload, so the legacy learned cache is never populated and the test passes identically on `main`.

I wrote the missing test to confirm: feed advertises `singleRider.isAvailable: true` on day one, POI drops the flag on day two. Passes on this branch, fails on `main` with `expected { waitTime: null } to be undefined`. That single test is the whole justification for the commit and it is not in the diff.

### 3. The `cacheVersion` collision with #293

`disneylandparis.ts:436`. #293 contains the byte-identical hunk taking `cacheVersion` 2 to 3, in a different block of the same file, so the whole thing merges clean and silent with no conflict. A host that warmed `:v3` under whichever PR landed first then serves that shape for the remaining 12h after the second lands, missing either tags or showtime durations depending on order.

Whichever of #293 and #295 lands second needs `cacheVersion: 4`. One thing that is not a problem and is worth knowing: the `@http` layer is safe on its own, because `generateCacheKey()` at `src/http.ts:360-369` includes the request body and both PRs change the GraphQL query text. `@cache` is the only exposure.

## Worth a decision: the single-rider swap trades a sticky retirement for a daily flap

`disneylandparis.ts:1071`. Replacing the learned cache with the POI flag is the right direction and today the two sources agree exactly, 9 of 9 over the published set. The one feed-only id, `P2YA03`, has no POI record and is not a published entity, so nothing is lost right now.

The regression is on the other transition. When Disney **adds** single rider, the wait feed knows and the 12h POI cache does not:

```
awake  -> {"STANDBY":{"waitTime":30},"SINGLE_RIDER":{"waitTime":15}}
asleep -> {"STANDBY":{"waitTime":null}}                      <- SINGLE_RIDER gone
```

On `main` both states carry the queue, because the learned cache bridged the gap. `@cache` and `fetchPOI`'s `cacheSeconds` are both 43200, so worst case is roughly 24h of layered staleness with the queue appearing and vanishing each day until POI catches up.

This repo's rule is to persist only the permissive value for derived booleans that gate emissions, precisely so a cached FALSE cannot become sticky. The removed cache was that permissive-only store and nothing replaced it. Keeping both properties is straightforward: OR the POI flag with a short-lived permissive cache, 24 to 48 hours rather than 30 days, fed from `wt.singleRider?.isAvailable === true`. Retirements then clear within a day or two and additions never flap.

Low frequency, so I am not treating it as blocking on its own, but I would rather not ship the flap knowingly.

## Smaller notes

- **`tags.test.ts:57-61` proves nothing.** `buildTags` is called unconditionally for every POI type at `:808` with no entity-type guard. The test passes only because the stubs withhold the fields. Give them the fields and the guarantee collapses:

  ```
  P1AR06 RESTAURANT [ 'MINIMUM_HEIGHT', 'SINGLE_RIDER', 'PAID_RETURN_TIME', 'ONRIDE_PHOTO' ]
  P1GS00 SHOW       [ 'SINGLE_RIDER', 'MAY_GET_WET' ]
  ```

  The docstring at `:672-673` ("shows and restaurants come back with an empty list") is a claim about the GraphQL query, not about the code. Latent only, since no non-Attraction POI category carries these fields today. Either add the guard or reword the docstring and the test name.
- **`TagBuilder.minimumHeightUnaccompanied()` exists.** If a two-height attraction is ever an accompanied/unaccompanied pair, first-match silently drops one, and `tags.test.ts:113-122` currently enshrines feed-order dependence as correct behaviour.
- **The comment block at `:1003-1004`** is grammatical but is now a verbatim duplicate of `:1006-1007`, and the paragraph explaining where the SINGLE_RIDER baseline comes from was deleted while the code 65 lines down still emits one. Tidy-up, not a defect. No dead import: `CacheLib` is still used at `:1009` and `:1016`.
- **Counts in the description are slightly off.** 54 tests are new; `visibility.test.ts` contributes the other 2 of the claimed 56. I count roughly 30 hostile-payload cases across the two files rather than 92. The substance is there, the number is not reproducible from the diff.
- The tests do drive the public path, with `getPOIData` as the only seam. Nothing asserts against `buildTags` or `parseHeightCm` directly, and the `expect(target).toBeDefined()` guard at `tags.test.ts:43` correctly stops the `toEqual([])` family passing on a vanished entity.

Worth recording for the history: this PR is purely additive in production. `main`'s `fetchPOI` requested no `height` or `physicalConsiderations` field at all, so the old `parseHeightCm` and its `poi.minimumHeight` branch were dead against the real feed and DLP published zero tags. The old localised-string parser would have produced `{height: 2000, unit: 'cm'}` from `"1,20 m"`; it never shipped only because the field was never fetched.

BeyondVertical:
All three blocking findings are fixed in 37eeac3e.

**1. Height band** — `parseHeightCm` now requires 40 to 200cm, so `1_20cm` and friends are rejected instead of rounding down to a published 1cm. Your five table rows are in the rejection suite, plus `39cm`/`201cm`, and the bounds (`40cm`, `200cm`, `2m`) are accepted-case tests. Verified live: heights still span exactly 81 to 140cm, 23 tagged of 129, all `ATTRACTION`.

**2. Retirement test** — the case the commit was named after is now in `singleRider.test.ts`: feed advertises `isAvailable: true` on day one, POI drops the flag, sighting lapses, queue gone.

**3. cacheVersion** — `getPOIData` moves to 4 on this branch. That turns the silent identical 2→3 merge with #293 into a visible conflict: if #293 lands first the merge resolves to 4, if this lands first #293's 3 conflicts and gets resolved by hand. Either order ends at 4 with no shared `:v3` key between the two shapes.

**On the flap** — taken, not shipped knowingly. The baseline now ORs the POI facet with wait-feed sightings from the last 48h, stored as per-id timestamps rather than a re-seeded set, so retirements actually lapse instead of living forever. Both transitions are under test: a fresh addition the POI cache does not carry yet keeps its queue (fails on the previous commit), and a retired queue disappears once the sighting lapses (via fake timers). 10 of the new tests fail against the pre-fix source.

Smaller notes:

- `buildTags` is now guarded on entity type, and the show/restaurant test hands them the full field set — your `P1AR06`/`P1GS00` demonstration is what it asserts against now. Docstring reworded to match.
- The duplicate comment block above the baseline section is deduped, and the single-rider baseline has its explanation back next to the new sighting cache.
- Description counts corrected: 67 tests now, and the unreproducible "92 payloads" figure is gone.
- `minimumHeightUnaccompanied`: leaving first-match for now since zero live attractions carry two height facets; happy to revisit if a pair ever appears.

Live check after the change: `SINGLE_RIDER` tag vs `queue.SINGLE_RIDER` still 9 = 9, `PAID_RETURN_TIME` 19 = 19, nothing in either direction only.

cubehouse (approved):
All three blocking findings are addressed. Approving.

Measured against the live feed on 2026-08-14 between 18:28 and 18:40 Paris, both parks open.

## The three blockers

**1. `parseHeightCm` bound, fixed.** The 40 to 200cm band at `:56-57` replaces the `cm > 0` lower bound, so `1_20cm` and `0_81cm` no longer round down to a published 1cm minimum.

I checked the narrowing costs nothing on real data. All 10 published `MINIMUM_HEIGHT` tags come back inside the band:

```
P1DA03 81   P2XA07 81   P1DA09 102  P2ZA02 102  P2XA03 107
P2XA06 120  P1DA08 120  P1AA02 140  + 2 more, all 81 to 140
```

81 to 140cm sits comfortably inside 40 to 200 with room either side, so nothing published today is at risk of being clipped.

**2. The retirement test exists now.** `singleRider.test.ts` covers the transition the commit is named after, and the addition case as well.

**3. `cacheVersion` collision resolved.** 4 here, so #293's 3 can no longer merge silently. Whichever of the two lands second keeps 4 or goes higher.

## The single-rider flap

The 48h permissive-only sightings cache at `:1023-1039` is the right shape. It stores only ids seen with `singleRider.isAvailable === true`, each with its own timestamp rather than being re-seeded, so an addition never flaps while the 12h POI cache lags and a retirement lapses within two days instead of never. That keeps the "persist only the permissive value" rule intact, which the plain POI-flag swap had dropped.

Live: 9 `SINGLE_RIDER` rows, matching the POI facet count exactly.

## Tag gating

`:814` gates `buildTags` on `entityType === 'ATTRACTION'`, and the test hands the stubs the fields rather than withholding them, so the guarantee is now about the code instead of about the fixture. Verified live: 129 entities, 23 tagged, and every tagged entity is an `ATTRACTION`.

## Merge note

This branch is behind `main` and predates #292 and #301, hence 129 entities here against 131 on `main`. Worth a rebase before merge so the numbers reconcile, though nothing in the diff depends on it.

BeyondVertical:
I see you rebased it yourself. I had rebased independently before fetching and landed on a byte-identical tree, so the two conflict resolutions agree: `TIME_OF_DAY` and the `P1G103` `attractionType` block kept from `main`, `buildTags` replacing the inline tag code, and the height band superseding the old `MAX_HEIGHT_CM = 300`.

Re-measured on the rebased head, so the numbers reconcile as you wanted:

```
entities        131          (was 129 pre-rebase, matches main)
tagged           23          all ATTRACTION
MINIMUM_HEIGHT   10 rides    81 to 140cm
PAID_RETURN_TIME 19   SINGLE_RIDER 9   ONRIDE_PHOTO 7
UNSUITABLE_PREGNANT 5  MAY_GET_WET 4
```

Full suite 1841 passing across 79 files, `tsc --noEmit` clean, harness 4/4.

One thing not comparable in that run: I measured at 23:05 Paris with the parks shut and after `--clear-cache`, so the 48h sightings store is cold and only one live row carries a `SINGLE_RIDER` queue. That is the cold-start row from your first review (`cold cache, feed asleep -> 0`) rather than a change in behaviour — the tag side is POI-derived and unaffected.

PR #296, opened 2026-08-12
fix(dlp): publish only schedules that can be attached to something

BeyondVertical:
## Summary

Three fixes in the DLP schedule and filter code, none of which changes an entity or a live-data row.

| | before | after |
|---|---|---|
| Schedule rows | 133 | **95** |
| Schedule days | 14,707 | **8,494** |
| Rows pointing at a non-existent entity | 38 | **0** |
| Entities | 129 | 129 |
| Live rows | 68 | 68 |
| DLP tests | 2 | **39** |

(Counts as measured at the time; the absolute day totals roll with the 60-day window and are not reproducible to the digit.)

## Schedules for entities that do not exist

`buildSchedules` emitted a row for everything the schedule feed returned. That feed covers far more than the POI set this destination publishes, so 38 of 133 rows — **6,213 of 14,707 days, 42% of the output** — pointed at ids no consumer can look up:

- 21 hotel and Disney Village restaurants (`H01R01` … `H07R01`, `D01R*`), which sit outside P1/P2 and are correctly not entities
- 14 character meets, whose `subType` is not in `SHOW_SUBTYPES`
- 2 railroad stations and `P1DA13` *Mickey's PhilharMagic*, dropped by the visibility filter
- the rest, records the ignore list covers

`buildEntityList` is now the authority on what may carry a schedule, rather than a second filter that can drift from it. Both parks keep all 120 days each, including every one of the 120 `EXTRA_HOURS` entries.

Worth saying plainly: most of today's excess is other work in flight, not a separate defect. Once #292's railroad stations and #293's character meets become entities, the same filter leaves **22 rows and 1,258 days**, almost all of it hotel dining. The headline number shrinks accordingly depending on merge order.

## One unusable row no longer costs every schedule

The feed's times were parsed with no validation, and `constructDateTime` throws on anything it cannot read — out of `buildSchedules` entirely, so all 95 entities lose all 60 days over one bad row. Fuzzing the public path, **14 of 19 shapes threw**, including the realistic case of one good row next to one broken one:

```
one good + one broken     THROW  Invalid time value      →  OK  rows=1 days=60
startTime = "9:30:00"     THROW  Invalid time value      →  OK  row skipped
schedules = [null]        THROW  reading 'status'        →  OK  row skipped
schedules = 42            THROW  not iterable            →  OK  entity skipped
activities = [null]       THROW  reading 'schedules'     →  OK  entity skipped
```

The time pattern is deliberately strict — two-digit hour, real minute and second ranges — because `9:30:00` reaches `constructDateTime` and throws exactly like `abc` does. Across **15,681 rows** in the live 60-day window it rejects none: every published time is `HH:MM:SS` with seconds of either `00` or `59`. The `:59` values are the 23:59 refurbishment placeholders, which is why the string is passed through untouched rather than truncated to `HH:MM`.

## A hide rule that has never matched anything

`HIDE_RULES` carried `Hide from Mobile App`. That string appears on **0 of 584** POI records — the value Disney emits is `Hide from the Mobile App`, as you noted reviewing #292.

Correcting the spelling is the obvious move and I think it is the wrong one. The 44 records wearing the real flag are guest-service articles (*Free WiFi*, *Parking Fees*, *Pets*, *Tax refunds*, *Expectant mothers*), day trips to Fontainebleau, Versailles and Provins, two off-site attractions, a third-party hotel and the Disney Village web radio. Content pages, not venues. None of them sits in a category and park this destination publishes, so the correction is measurably a no-op — 126 entities either way, nothing lost, nothing gained — while quietly turning "hidden in the app" into a reason to drop a ride. That is the opposite of where #292 landed for the two railroad stations.

So the dead rule is removed rather than repaired, and `visibility.test.ts` now pins a show wearing the flag as published, so a future spelling fix has to argue with a red test first.

## Merge notes

Touches `buildSchedules`, the `HIDE_RULES` constant and one new constant after `SHOW_SUBTYPES`. #293 and #295 both add a constant in that same spot, so expect a one-hunk conflict with whichever lands first; there is no conflict in `buildSchedules` itself. If #294 lands first, its `TIME_OF_DAY` is the looser `/^\d{2}:\d{2}/` — keep this stricter one, which satisfies both call sites.

Happy to split this into separate PRs if you would rather review them apart.

## Test plan

- [x] `npm test src/parks/dlp` — 39 tests across 2 files
- [x] `npm run dev -- disneylandparis --clear-cache` — 4/4, 129 entities and 68 live rows unchanged, 95 schedules
- [x] Full suite: 1,723 passing; the two failures in `cache.test.ts` / `cachePragmas.test.ts` are pre-existing on `main` (SQLite WAL pragmas on Windows)
- [x] 19 malformed schedule payloads driven through the public `getSchedules()` path, none throwing

### DIFF
--- a/src/parks/dlp/__tests__/schedules.test.ts
+++ b/src/parks/dlp/__tests__/schedules.test.ts
@@ -1,0 +1,112 @@
+import {describe, it, expect, vi, beforeEach} from 'vitest';
+import {DisneylandParis} from '../disneylandparis.js';
+
+/**
+ * The schedule feed covers far more than the POI set the park publishes —
+ * hotel and Disney Village restaurants, character meets, and records the
+ * visibility filter drops. Only entities that actually get emitted may carry
+ * schedules.
+ */
+const OPERATING = {startTime: '09:30:00', endTime: '22:00:00', status: 'OPERATING'};
+
+function stubbedPark(activities: unknown): DisneylandParis {
+  const park = new DisneylandParis();
+
+  vi.spyOn(park as any, 'getPOIData').mockResolvedValue({
+    ThemePark: [{id: 'P1', name: 'Disneyland Park', type: 'ThemePark'}],
+    Attraction: [
+      {id: 'P1RA00', name: 'Big Thunder Mountain', type: 'Attraction', location: {id: 'P1'}},
+      {
+        id: 'P1DA13',
+        name: 'Mickey’s PhilharMagic',
+        type: 'Attraction',
+        location: {id: 'P1'},
+        hideFunctionality: 'Hide from the Service',
+      },
+      {id: 'P2AC00', name: 'Ignored', type: 'Attraction', location: {id: 'P2'}},
+    ],
+    Entertainment: [
+      {id: 'P1MG03', name: 'Meet Goofy', type: 'Entertainment', subType: 'Meet', location: {id: 'P1'}},
+    ],
+    Restaurant: [
+      {id: 'H01R01', name: 'La Table de Lumière', type: 'Restaurant', location: {id: 'H01'}},
+    ],
+  });
+  vi.spyOn(park as any, 'getScheduleForDate').mockResolvedValue(activities);
+
+  return park;
+}
+
+/** Ids that end up with a schedule, given what the schedule feed returned. */
+async function scheduledIds(activities: unknown): Promise<string[]> {
+  const schedules = await stubbedPark(activities).getSchedules();
+  return schedules.map((s) => s.id).sort();
+}
+
+const feedFor = (ids: string[], hours: unknown = OPERATING) =>
+  ids.map((id) => ({id, name: id, schedules: [hours]}));
+
+describe('DLP schedules', () => {
+  beforeEach(() => vi.restoreAllMocks());
+
+  // === Only published entities ===
+
+  it('keeps a schedule for a published attraction', async () => {
+    expect(await scheduledIds(feedFor(['P1RA00']))).toEqual(['P1RA00']);
+  });
+
+  it('keeps park schedules', async () => {
+    expect(await scheduledIds(feedFor(['P1']))).toEqual(['P1']);
+  });
+
+  it.each([
+    ['a hotel restaurant outside the parks', 'H01R01'],
+    ['an attraction the visibility filter drops', 'P1DA13'],
+    ['an entity on the ignore list', 'P2AC00'],
+    ['entertainment that is not a show subtype', 'P1MG03'],
+    ['an id absent from the POI feed entirely', 'D01R02'],
+  ])('drops the schedule for %s', async (_label, id) => {
+    expect(await scheduledIds(feedFor([id]))).toEqual([]);
+  });
+
+  it('keeps only the published ids when the feed mixes them', async () => {
+    expect(await scheduledIds(feedFor(['P1', 'P1RA00', 'H01R01', 'P1DA13', 'D01R02'])))
+      .toEqual(['P1', 'P1RA00']);
+  });
+
+  // === Schedule contents ===
+
+  it('publishes one entry per day of the 60-day window', async () => {
+    const schedules = await stubbedPark(feedFor(['P1RA00'])).getSchedules();
+    expect(schedules[0].schedule).toHaveLength(60);
+  });
+
+  it('maps EXTRA_MAGIC_HOURS and PERFORMANCE_TIME to their own types', async () => {
+    const magic = await stubbedPark(feedFor(['P1'], {
+      startTime: '08:30:00', endTime: '09:30:00', status: 'EXTRA_MAGIC_HOURS',
+    })).getSchedules();
+    expect(magic[0].schedule[0]).toMatchObject({type: 'EXTRA_HOURS', description: 'Extra Magic Hours'});
+
+    const show = await stubbedPark(feedFor(['P1RA00'], {
+      startTime: '21:00:00', endTime: '21:30:00', status: 'PERFORMANCE_TIME',
+    })).getSchedules();
+    expect(show[0].schedule[0]).toMatchObject({type: 'INFO', description: 'Performance Time'});
+  });
+
+  it.each(['REFURBISHMENT', 'CLOSED'])('skips %s days', async (status) => {
+    expect(await scheduledIds(feedFor(['P1RA00'], {...OPERATING, status}))).toEqual([]);
+  });
+
+  it('keeps the seconds the feed publishes', async () => {
+    const schedules = await stubbedPark(feedFor(['P1RA00'], {
+      startTime: '00:00:00', endTime: '23:59:00', status: 'OPERATING',
+    })).getSchedules();
+    expect(schedules[0].schedule[0].closingTime).toMatch(/T23:59:00/);
+  });
+
+  it('survives a schedule fetch that rejects', async () => {
+    const park = stubbedPark(null);
+    vi.spyOn(park as any, 'getScheduleForDate').mockRejectedValue(new Error('upstream down'));
+    await expect(park.getSchedules()).resolves.toEqual([]);
+  });
+});
--- a/src/parks/dlp/__tests__/schedules.test.ts
+++ b/src/parks/dlp/__tests__/schedules.test.ts
@@ -1,4 +1,4 @@
-import {describe, it, expect, vi, beforeEach} from 'vitest';
+import {describe, it, expect, vi, afterEach} from 'vitest';
 import {DisneylandParis} from '../disneylandparis.js';
 
 /**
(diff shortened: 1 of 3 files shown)


cubehouse (changes requested):
This is careful work and most of it verified exactly as described. One thing needs fixing before it lands.

What I confirmed against the live feed:

- **Rows 133 to 95, orphans 38 to 0.** Cross-checked `getSchedules()` output against `getEntities()`: 0 remaining orphans.
- **Every dropped day reconciles.** `OPERATING` 5,611 to 4,317, and the 1,294 lost are 21 hotel and Disney Village restaurants (1,198) plus `P1DA13` (60), `P1DA10` (6) and `P1NA16` (30). `INFO` 8,773 to 3,970, and the 4,803 lost are the 14 character-meet ids, to the day. Nothing unexplained.
- **Both parks intact**, 120 days each, and `EXTRA_HOURS` 120 to 120, zero lost.
- **The strict regex rejects nothing**: 0 of 15,477 live rows fail `TIME_OF_DAY` on either field, across all statuses.
- **"14 of 19 shapes threw before the fix" is exactly right.** I replayed all 19 payloads through `getSchedules()` against the pre-hardening source: threw 14 of 19. Post-PR: 0 of 19. The load-bearing one is a good row next to a broken one, which did cost every entity all 60 days.
- **The tests drive the public path**, stubbing only `getPOIData` and `getScheduleForDate`. Against pre-PR source the new suite is 19 failed, 16 passed.
- **Removing the `IGNORE_ENTITIES` check is safe**: every ignored id is already excluded by `filterPOIEntities` at `:606`, and the 0 orphan count proves nothing slips through.
- **The `Hide from Mobile App` removal is correct.** Independently confirmed the string appears 0 times across all 584 POI records; the value Disney emits is `Hide from the Mobile App` (44). Your argument for removing rather than repairing it is right.

Cost is a non-issue, measured: warm `buildEntityList()` repeats at 3.6 to 4.2ms against a `getSchedules()` that takes about 7.3 seconds, so roughly 0.05%.

## Blocking: a POI failure now empties the schedule feed

`disneylandparis.ts:1083`. Filtering on `buildEntityList()` couples schedules to POI availability, including the silent path. `getPOIData` at `:399-402` ends `return data?.data || {}`, so a GraphQL 200-with-errors body becomes `{}` with no exception, and `@cache({ttlSeconds: 43200})` pins that for 12 hours.

Measured by stubbing `getPOIData` and feeding the real 60-day schedule snapshot through the public `getSchedules()`:

| POI response | pre-PR (`main`) | this PR |
|---|---|---|
| healthy | 133 rows / 14,504 days | 95 / 8,407 |
| request throws | 133 / 14,504 | **rejects** `Error: POI upstream 503` |
| `{}` | 133 / 14,504 | **1 row / 120 days (P2 only)** |
| ThemePark records only | 133 / 14,504 | 2 rows / 240 days |

The `{}` row is the dangerous one because it needs no exception to reach. `buildEntityList` injects P2 when missing at `:710-718` but never P1, so the output becomes one park's hours and nothing else: Disneyland Park's operating hours, both parks' attractions, all restaurants and all shows silently losing their schedules for half a day. Before this change the same POI outage left schedules completely intact.

Cheap fix, treating an implausible entity list as "do not filter" rather than "filter everything out":

```ts
let publishedIds: Set<string> | null = null;
try {
  const entities = await this.buildEntityList();
  // 3 = destination + 2 injected parks; at or below that is a POI failure
  if (entities.length > 3) publishedIds = new Set(entities.map((e) => e.id));
} catch { /* POI unavailable: publish unfiltered rather than nothing */ }
...
if (publishedIds && !publishedIds.has(entity.id)) continue;
```

The suite fuzzes the schedule feed thoroughly but never stubs `getPOIData` to throw or return `{}`, which is the one path this regresses. A test asserting "POI unavailable does not empty the schedule feed" belongs in the same commit.

## Notes, none blocking

- **Consider `getEmittablePOIEntities()` rather than `buildEntityList()`.** The helper at `:662-672` exists precisely so "the wait-times / premier-access / vqueue / showtimes paths can't surface IDs that fall outside the published POI set", and `buildLiveData` uses it at `:819`. `buildEntityList` additionally runs `mapEntities`, height parsing and `TagBuilder` for 129 entities and mutates instance state (`this.showDurationMap.clear()` at `:734`, repopulated at `:790`). It is safe, since there is no `await` between the clear and the repopulation so no torn read is possible, but a `build*` that silently rewrites a field another `build*` reads is surprising. Note `getEmittablePOIEntities()` is not a drop-in: `mapEntityType` returns `undefined` for `category === 'ThemePark'`, so using it directly would delete both parks' 240 schedule days. A small `getPublishedEntityIds()` helper (emittable POI ids plus the park ids plus `dlp`) is the cleaner shape. To be clear on the contract question, this is not a Template Method violation: `src/destination.ts` and CLAUDE.md forbid overriding `get*()`, not calling `build*` from `build*`.
- **The optional-seconds group is dead and opens a compare hole.** Across all 15,477 live rows every time carries seconds (30,460 `:00` plus 494 `:59`, zero remainder, no mixed-length pairs), so `(:[0-5]\d)?` at `:61` never fires on real data. It does admit `HH:MM`, and `:1117` compares the raw strings, so `endTime '09:30'` against `startTime '09:30:00'` sorts as less-than and would fabricate a midnight rollover, turning a zero-length window into 24 hours. Theoretical today; either require seconds or normalise before comparing.
- **`schedules.test.ts:100-105` pins a shape the kept-row path never sees.** All 494 non-`:00` seconds values are `endTime: 23:59:59` and every one is on a `REFURBISHMENT` row, which `:1106` discards before the regex or `constructDateTime` is reached. So "seconds of either 00 or 59, passed through untouched" is true about the feed but vacuous about behaviour. Harmless as a contract test, but the commit message overstates it. Your reading that `:59` is the refurbishment placeholder is confirmed.
- **`visibility.test.ts:63-66` does not pin the change its commit makes.** All three visibility tests pass against pre-PR source, because `Hide from the Mobile App` was never in `HIDE_RULES` on either side. It is a guard against a future spelling fix, exactly as the commit says, but nothing goes red if someone re-adds the dead `Hide from Mobile App` string.
- **Third copy of a homegrown time validator.** `:61` here, `energylandia.ts:398`, and the looser one in #294, while `constructDateTime` has 169 call sites repo-wide and throws `RangeError` on anything it cannot parse. Third occurrence is the point at which this belongs in `src/datetime.ts` as `isTimeOfDay()` / `constructDateTimeSafe()` with its own tests, rather than a fourth local copy.
- `beforeEach(() => vi.restoreAllMocks())` at `schedules.test.ts:50` and `visibility.test.ts:48` should be `afterEach`. Harmless here since every test builds a fresh instance.
- Minor: this touches two test files, not three, and the absolute day counts in the commit message are unreproducible by design, since the 60-day window rolls past Paris midnight. My run gave 14,504 to 8,407 against your 14,707 to 8,494, and the harness a few minutes later reported 8,382. Not an error, just worth not pinning in prose.
- `P1DA13` is the one dropped row worth a second look, and dropping it is right here: the schedule feed publishes it `OPERATING 12:30-17:30` for all 60 days while the wait feed reports `REFURBISHMENT`. That is being handled separately in #297.

## Beyond this PR

`src/testRunner.ts:247-268` validates schedule structure but never cross-checks `schedule.id` against the entity list. A generic assertion there would have caught these 38 orphans automatically and would surface the same defect across the other 74 parks. Worth its own issue; I will raise it.

Gates all green on the branch: DLP tests 35 passed, full suite 75 files and 1,725 tests passed, `tsc --noEmit` clean, harness 4/4.

BeyondVertical:
The blocking finding is fixed in 3e56d20f, using the `getPublishedEntityIds()` shape from your notes rather than the inline threshold.

**POI failure no longer empties the feed.** The helper returns the emittable POI ids plus `P1`/`P2`/`dlp`, and null when `getPOIData` throws or yields no emittable record — null meaning "do not filter". All three failure modes from your table are under test through the public `getSchedules()` (reject, `{}`, ThemePark-records-only) and all three fail against the pre-fix source. This also drops the `buildEntityList()` call, so `buildSchedules` no longer runs `mapEntities`/`TagBuilder` or touches `showDurationMap` — your ThemePark caveat is what the hardcoded park ids in the helper are for.

From the notes:

- The midnight-rollover comparison is on `HH:MM` for both sides, so a mixed `HH:MM` / `HH:MM:SS` pair can't fabricate a 24h window. I kept the optional-seconds group in the regex rather than requiring seconds — with the compare hole closed it only widens what's accepted, not what's misread.
- `afterEach` in both test files.
- Description corrected: two test files, current test count, and the absolute day totals are now marked as measured-at-time rather than pinned.
- On the third homegrown time validator: agreed it belongs in `src/datetime.ts`, but #294 and this branch both touch the same constant and will conflict at merge anyway — I'd rather do the `isTimeOfDay()` extraction as a follow-up across all three call sites once both land, instead of adding a cross-branch dependency now. Happy to file it.

`main` (with #292) is merged in. Re-measured live after the change: 97 rows / 8,290 days (the two railroad stations from #292 now carry rows), 0 orphans, both parks at 120 days.

cubehouse (approved):
Blocking finding fixed, and all three failure modes verified rather than reasoned about. Approving.

Measured against the live feed on 2026-08-14 between 18:28 and 18:40 Paris, both parks open.

## The POI-failure path

`getPublishedEntityIds()` at `:679` returns `null` rather than an empty set when POI is unavailable, and `:1156` treats `null` as "do not filter". Driven through the public `getSchedules()` with the real schedule feed behind it:

| POI response | schedule rows |
|---|---|
| healthy | 97 (from 135), 8,290 days, **0 orphans** |
| `{}` | 135, unfiltered |
| throws | 135, unfiltered, no rejection |

That is exactly the behaviour asked for: a silent GraphQL 200-with-errors body no longer empties the feed for 12 hours, and an exception no longer rejects the whole promise.

Using `getEmittablePOIEntities()` rather than `buildEntityList()` is better than the `> 3` threshold I suggested. It sidesteps the `showDurationMap` mutation, and because the injected parks are added explicitly afterwards, `length === 0` is an exact test for "POI gave us nothing" instead of a heuristic.

The `HH:MM` normalisation on the midnight-rollover compare at `:1143` closes the mixed-length hole, and mock restoration is on `afterEach`.

## Blocking on merge order, not on this diff

**This PR and #297 auto-merge with no conflict in `disneylandparis.ts`, and the result silently breaks #297.** Only `visibility.test.ts` conflicts. In the merged file:

```
1156:  if (publishedIds && !publishedIds.has(entity.id)) continue;   // this PR
1160:  const scheduleId = ID_ALIASES[entity.id] ?? entity.id;        // #297
```

`P1DA13` is hidden and never in `publishedIds`, so its 60 folded `OPERATING` rows are dropped at 1156 before the alias at 1160 can move them onto `P1G103`. #297's schedule half stops working and git does it quietly.

One line reordered fixes it:

```ts
const scheduleId = ID_ALIASES[entity.id] ?? entity.id;
if (publishedIds && !publishedIds.has(scheduleId)) continue;
```

**Land #297 first, then this.** #297's `attaches the twin's schedule rows to P1G103` test goes red if the rebase gets the order wrong, which is the guard you want. The reverse order leaves it to whoever hand-resolves `visibility.test.ts`.

## Other merge note

`TIME_OF_DAY` is defined differently here and in #294. Those two conflict loudly on a duplicate `const`, so nothing slips through, but keep #294's version: it also accepts `24:00` as a midnight close, and `constructDateTime` handles `24:00:00` correctly (`new Date('2026-08-14T24:00:00Z')` gives `2026-08-15T00:00:00.000Z`, and the convergence loop lands on Paris midnight).

PR #297, opened 2026-08-12
feat(dlp): surface Mickey's PhilharMagic

BeyondVertical:
Mickey's PhilharMagic ran without a published standby wait: its wait-feed row belongs to `P1DA13`, an Attraction record flagged `Hide from the Service` and filtered out of the POI list. It is not retired — the wait feed carries it alongside the rest of the ride estate, and the activity-schedule feed reports it OPERATING on every day of the window.

Review surfaced why simply excepting `P1DA13` was wrong: POI holds **two records for the same theatre**, and the other one is already published.

```
[Attraction]    P1DA13 | Mickey's PhilharMagic | hide="Hide from the Service" | wait feed, derived OPERATING schedule
[Entertainment] P1G103 | Mickey's PhilharMagic | no flag                      | showtimes (live on the wiki today)
```

`P1DA13`'s 60-day OPERATING window is exactly the first-to-last performance time of `P1G103` and tracks the timetable when it shifts — a derived restatement. Surfacing it would have created a name-level duplicate with the venue's data split in half.

**Approach:** keep the published id and fold the twin onto it.

- `P1G103` is emitted as an `ATTRACTION` via an id-keyed override ahead of the category mapping — the classification Walt Disney World, Disneyland Resort and Hong Kong use — with `attractionType: SHOW` set explicitly so `destination.ts` cannot default a 4D cinema to `RIDE`.
- `ID_ALIASES = {P1DA13: 'P1G103'}` folds the twin's rows onto the published id in the wait-times loop, the queue-bearing history and `buildSchedules`.
- `P1DA13` stays filtered; `VISIBILITY_EXCEPTIONS` and the hide rules are untouched.

**Effect:** entities stay at 131, and the published entity — stable id, stable UUID — now carries both halves. Verified live: `P1G103` reports `OPERATING` with a `STANDBY` wait and 11 showtimes, its schedule holds the performance times plus the twin's 60 OPERATING days, `P1DA13` appears nowhere, and orphaned schedule ids go 36 → 35 (the rest is #296's territory).

`P1JF00` Gardens of Wonder — the only other attraction carrying the flag, absent from every feed — remains the control: it stays filtered, and the test goes red if anyone widens the rule.

## Test plan

- [x] `npm test src/parks/dlp` — cardinality (exactly one PhilharMagic), `ATTRACTION`/`SHOW` type pair, folded wait time next to surviving showtimes, schedule attachment; 3 of 6 fail against the previous approach
- [x] `npx tsc --noEmit`
- [x] Verified against the live API (entities, live data, schedules — numbers above)

### DIFF
--- a/src/parks/dlp/__tests__/visibility.test.ts
+++ b/src/parks/dlp/__tests__/visibility.test.ts
@@ -1,11 +1,12 @@
-import {describe, it, expect, vi, beforeEach} from 'vitest';
+import {describe, it, expect, vi, afterEach} from 'vitest';
 import {DisneylandParis} from '../disneylandparis.js';
 
 /**
- * A hidden POI is normally dropped by filterPOIEntities (HIDE_RULES). Entities
- * listed in VISIBILITY_EXCEPTIONS bypass that. P1GS93 ("Live Your Story") is a
- * real Castle Stage show Disney flags "Hide from the Service", so it must be
- * force-surfaced while other hidden shows stay filtered.
+ * A hidden POI is normally dropped by filterPOIEntities (HIDE_RULES); entities
+ * in VISIBILITY_EXCEPTIONS bypass that. Mickey's PhilharMagic exists twice in
+ * POI — published Entertainment record P1G103 and hidden Attraction twin
+ * P1DA13 — and must surface exactly once, as P1G103, with the twin's live
+ * and schedule data folded onto it.
  */
 function stubbedPark(): DisneylandParis {
   const park = new DisneylandParis();
@@ -35,6 +36,7 @@ function stubbedPark(): DisneylandParis {
         hideFunctionality: 'Hide from Web List + Mobile App',
       },
       {
+        // The hidden PhilharMagic twin: stays filtered, its data is aliased.
         id: 'P1DA13',
         name: 'Mickey’s PhilharMagic',
         type: 'Attraction',
@@ -69,13 +71,27 @@ function stubbedPark(): DisneylandParis {
         location: {id: 'P1'},
         hideFunctionality: 'Hide from Web List + Mobile App',
       },
+      {
+        // The published PhilharMagic record.
+        id: 'P1G103',
+        name: 'Mickey’s PhilharMagic',
+        type: 'Entertainment',
+        subType: 'Stage Show',
+        location: {id: 'P1'},
+      },
     ],
   });
   return park;
 }
 
+function stubLiveFeeds(park: DisneylandParis, waitTimes: unknown[] = []): void {
+  vi.spyOn(park as any, 'getWaitTimes').mockResolvedValue(waitTimes);
+  vi.spyOn(park as any, 'getPremierAccess').mockResolvedValue([]);
+  vi.spyOn(park as any, 'getVirtualQueueData').mockResolvedValue([]);
+}
+
 describe('DLP visibility exceptions', () => {
-  beforeEach(() => vi.restoreAllMocks());
+  afterEach(() => vi.restoreAllMocks());
 
   it('surfaces railroad stations flagged Hide from Web List + Mobile App', async () => {
     const entities = await stubbedPark().getEntities();
@@ -97,12 +113,45 @@ describe('DLP visibility exceptions', () => {
     expect((lys as any)?.parkId).toBe('P1');
   });
 
-  it('surfaces P1DA13 (Mickey’s PhilharMagic) despite its "Hide from the Service" flag', async () => {
+  it('publishes exactly one PhilharMagic: P1G103 as a show-type attraction', async () => {
     const entities = await stubbedPark().getEntities();
-    const philharMagic = entities.find((e) => e.id === 'P1DA13');
-    expect(philharMagic).toBeDefined();
-    expect(philharMagic?.entityType).toBe('ATTRACTION');
-    expect((philharMagic as any)?.parkId).toBe('P1');
+    const matches = entities.filter((e) => /philharmagic/i.test(String(e.name)));
+    expect(matches).toHaveLength(1);
+    expect(matches[0].id).toBe('P1G103');
+    expect(matches[0].entityType).toBe('ATTRACTION');
+    expect((matches[0] as any).attractionType).toBe('SHOW');
+    expect((matches[0] as any).parkId).toBe('P1');
+  });
+
+  it('folds the hidden twin\'s wait time onto P1G103 and keeps its showtimes', async () => {
+    const park = stubbedPark();
+    stubLiveFeeds(park, [
+      {entityId: 'P1DA13', type: 'Attraction', status: 'OPERATING', postedWaitMinutes: 15},
+    ]);
+    vi.spyOn(park as any, 'getScheduleForDate').mockImplementation(async (date) => [
+      {id: 'P1G103', schedules: [
+        {date, startTime: '12:30:00', endTime: '12:30:00', status: 'PERFORMANCE_TIME'},
+      ]},
+    ]);
+
+    const live = await park.getLiveData();
+    expect(live.find((l) => l.id === 'P1DA13')).toBeUndefined();
+    const row = live.find((l) => l.id === 'P1G103');
+    expect(row?.queue?.STANDBY).toEqual({waitTime: 15});
+    expect(row?.showtimes).toHaveLength(1);
+  });
+
+  it('attaches the twin\'s schedule rows to P1G103 instead of orphaning them', async () => {
+    const park = stubbedPark();
+    vi.spyOn(park as any, 'getScheduleForDate').mockImplementation(async (date) => [
+      {id: 'P1DA13', schedules: [
+        {date, startTime: '12:30:00', endTime: '17:30:00', status: 'OPERATING'},
+      ]},
+    ]);
+
+    const schedules = await park.getSchedules();
+    expect(schedules.find((s) => s.id === 'P1DA13')).toBeUndefined();
+    expect(schedules.find((s) => s.id === 'P1G103')?.schedule).toHaveLength(60);
   });
 
   it('still drops other retirement-flagged POIs not in the exception set', async () => {
--- a/src/parks/dlp/__tests__/visibility.test.ts
+++ b/src/parks/dlp/__tests__/visibility.test.ts
@@ -34,6 +34,22 @@ function stubbedPark(): DisneylandParis {
         location: {id: 'P1'},
         hideFunctionality: 'Hide from Web List + Mobile App',
       },
+      {
+        id: 'P1DA13',
+        name: 'Mickey’s PhilharMagic',
+        type: 'Attraction',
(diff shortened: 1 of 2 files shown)


cubehouse (changes requested):
The evidence in the description is sound and I verified all of it. The problem is a record the description does not mention.

## There are two POI records for this theatre, and one is already published

```
[Attraction]    P1DA13 | Mickey's PhilharMagic | subType=""           | hide="Hide from the Service" | loc=P1
[Entertainment] P1G103 | Mickey's PhilharMagic | subType="Stage Show" | hide=""                      | loc=P1
```

`P1G103` has no hide flag, is emitted today as a `SHOW`, and is live on the wiki right now under a stable UUID. So PhilharMagic is not missing. What is missing is its standby wait time.

On this branch, `getEntities()` returns 132 and both records are emitted:

```
P1DA13  entityType=ATTRACTION  parentId=P1  attractionType=RIDE   (defaulted by destination.ts:901)
P1G103  entityType=SHOW        parentId=P1
```

A duplicate check over parentId plus lowercased name shows the new collision:

```
P1|mickey's philharmagic -> P1DA13/ATTRACTION , P1G103/SHOW      <- introduced by this PR
P1|disneyland railroad main street station -> P1MA01 , P1MA05    <- pre-existing on main
```

And the venue's data splits in half rather than combining:

```
P1DA13 {"id":"P1DA13","status":"OPERATING","queue":{"STANDBY":{"waitTime":15}}}           <- wait, no showtimes
P1G103 {"id":"P1G103","status":"OPERATING","showtimes":[ 11 performances 12:30..17:30 ]}  <- showtimes, no queue
```

```
P1DA13   60 rows  type=OPERATING             e.g. 2026-08-13 12:30 -> 17:30
P1G103  539 rows  type=INFO Performance Time  e.g. 2026-08-13 12:30, 13:00, 13:30 ...
```

The clincher is that over 60 consecutive days `P1DA13`'s OPERATING window is exactly the first-to-last performance time of `P1G103`, and it tracks the timetable when it shifts:

```
2026-08-13  P1DA13 OPERATING 12:30-17:30   P1G103 12:30 13:00 ... 17:00 17:30
2026-09-07  P1DA13 OPERATING 11:00-16:00   P1G103 11:00 11:30 ... 15:30 16:00
```

`P1DA13`'s schedule is a derived restatement of `P1G103`'s showtimes. Same venue, and the two entities each hold half of its data.

## What to do instead

This should be one entity carrying both the showtimes and the standby wait. You are right that it belongs in the ride estate rather than the entertainment estate: it sits in the wait feed with the full ride shape, and Walt Disney World, Disneyland Resort and Hong Kong all classify it as an attraction. The part to change is which record survives.

**Keep the published id `P1G103`, change its type to `ATTRACTION`, and alias `P1DA13`'s live data onto it.** I have confirmed on our side that changing `entityType` on an already-published entity preserves its UUID, so this costs nothing downstream. `P1DA13` stays filtered and the hide rules stay untouched.

Concretely:

1. **Drop the `P1DA13` line from `VISIBILITY_EXCEPTIONS`.** The visibility filter is not the right lever here.
2. **Emit `P1G103` as `ATTRACTION`.** Its category is `Entertainment` with `subType: 'Stage Show'`, so `mapEntityType` needs a narrow id-keyed override ahead of the subtype lookup.
3. **Set `attractionType` explicitly.** `destination.ts:901` defaults any `ATTRACTION` without one to `RIDE`, which would ship a 4D cinema as a ride. `AttractionTypeEnum` includes `SHOW`, which describes it exactly and is what I would use. If you would rather match what the wiki currently shows for the Walt Disney World version, that is `UNKNOWN`; either is defensible, just do not let it default.
4. **Alias the wait time.** A `LIVE_ID_ALIASES = {P1DA13: 'P1G103'}` applied in the wait-times loop at `disneylandparis.ts:836` before `getOrCreate` is enough.
5. **Handle `P1DA13`'s schedule rows**, either aliased or skipped, so its 60 redundant OPERATING rows stop being orphaned.
6. **Showtimes survive the flip.** They are not gated on `entityType` anywhere in `src/destination.ts`, so `P1G103` keeps its 11 daily performances as an attraction. Worth an assertion rather than an assumption.

The end state is one entity, stable id and UUID, carrying showtimes and a live standby queue, classified consistently with the other Disney resorts.

## The test

The stub at `visibility.test.ts:37-52` mirrors the real POI record faithfully, and the test does fail against pre-PR source (`expected undefined to be defined` at `:103`), so it asserts something real. The `P1JF00` control at `:112` is a genuine guard that fails the moment anyone widens `HIDE_RULES` instead of adding an individual id. Both are worth keeping in whatever shape this takes next.

The gap is that the stub omits `P1G103` entirely, so the suite is structurally blind to the duplication. The test that should exist, "exactly one entity named Mickey's PhilharMagic in P1", would fail on this branch today. Add `P1G103` to the stub and assert cardinality, its `entityType` and `attractionType`, and that it carries both a `STANDBY` queue and showtimes.

Nit: the file's header docblock at `:4-9` still describes only `P1GS93` and was not updated for the new cases.

## Claims checked

| Claim | Result |
|---|---|
| Schedule feed reports P1DA13 OPERATING on every one of the next 30 days | Verified, and on all 60 I pulled |
| Wait feed carries it alongside the rest of the ride estate | Verified, a normal `type:"Attraction"` row |
| ...reporting OPERATING today | Not reproducible at my run time. At 23:58 Paris it read REFURBISHMENT, but so did 67 of 68 rows, since DLP parks the whole estate overnight. No evidence against it |
| Only two attractions carry "Hide from the Service" | Verified, exactly `P1JF00` and `P1DA13` |
| P1JF00 appears in no live feed | Verified, absent from the wait feed on both pulls, absent from all 60 days of schedules, and its POI record carries 0 schedules |
| Entities 131 to 132 | Verified exactly |

One factual note: the description calls `Hide from the Service` "the flag Disney otherwise uses to mark retired records". 24 P1/P2 records carry it including live ones, such as `P1GS10` Disney's Christmas Parade, `P1NR03` Fantasia Gelati and `P1DR04` Lounge Lightspeed. It is not a retirement marker, and the existing `P1GS93` comment already hedges this. The retirement marker is `Hide from Web List + Mobile App`, which is the one #292 dealt with.

Also worth knowing: `P1DA13`'s 60 OPERATING rows are already emitted on `main` as orphans, since `buildSchedules` applies no visibility filter. Orphan schedule ids go 36 to 35 with this PR, so "no orphaned rows introduced" is true, but the pre-PR state was one orphan rather than zero. #296 removes the rest.

Thanks for digging this one out. The wait time genuinely was missing and finding it was the hard part; it is only the plumbing that needs redirecting.

BeyondVertical:
Reworked in ecdf43cb along the lines you laid out — the visibility exception is gone and the plumbing redirects instead.

All six steps, as implemented:

1. `P1DA13` is out of `VISIBILITY_EXCEPTIONS`; hide rules untouched.
2. `P1G103` emits as `ATTRACTION` via an id-keyed `ENTITY_TYPE_OVERRIDES` checked ahead of the category mapping.
3. `attractionType: 'SHOW'` is set explicitly — took your recommendation over `UNKNOWN`.
4. `ID_ALIASES = {P1DA13: 'P1G103'}` applies in the wait-times loop before `getOrCreate`, and also in the queue-bearing history so the overnight STANDBY baseline lands on the published id rather than evaporating with the filtered twin.
5. The twin's schedule rows are aliased rather than skipped — the venue's operating window is real data, so `P1G103` now carries the 60 OPERATING days alongside its performance times, and the orphan disappears (36 → 35 on this branch; #296 handles the rest).
6. Showtimes surviving the flip is asserted, not assumed: the live-data test pins a `STANDBY` wait and showtimes on the same row.

The test gap you called out is closed the way you described: the stub now contains both records, and the suite asserts cardinality ("exactly one entity named Mickey's PhilharMagic in P1"), the `ATTRACTION`/`SHOW` pair, and both data halves on `P1G103`. The `P1JF00` control stays. 3 of the 6 tests fail against the previous approach. Docblock updated for the new shape.

Verified live: entities stay at 131, `P1G103` reports OPERATING with a STANDBY wait and 11 showtimes, its schedule holds 528 performance-time days plus the twin's 60 OPERATING days, and `P1DA13` appears in no output.

The description is rewritten around the two-record situation, and the "retirement marker" claim about `Hide from the Service` is gone — your list of live carriers settles that.

cubehouse (approved):
Reworked exactly as asked, and it works end to end against the live feed. Approving.

Measured on 2026-08-14 between 18:28 and 18:40 Paris, both parks open, with `main` measured in the same window as a baseline so feed drift cannot flatter the result.

```
main     131 entities   P1G103  SHOW              no queue, 11 showtimes
                        P1DA13  not an entity     schedule orphan, 60 OPERATING days
                        schedules: 135 rows / 38 orphans
                        P1G103 schedule: 528 INFO

pr/297   131 entities   P1G103  ATTRACTION / SHOW  STANDBY queue + 11 showtimes
                        P1DA13  absent from live and schedules
                        schedules: 134 rows / 37 orphans
                        P1G103 schedule: 528 INFO + 60 OPERATING
```

Exactly one PhilharMagic, on the already-published id, carrying both halves of the venue. Entity count unchanged, one orphan retired, showtimes survive the `entityType` flip as predicted. `attractionType` is set explicitly so `destination.ts:901` cannot default a 4D cinema to `RIDE`.

The `ID_ALIASES` placement is right in all three spots that needed it: the wait-times loop at `:859`, the queue-bearing history at `:1021`, and `buildSchedules`. The tests assert cardinality rather than mere presence, which is what makes them able to fail on the previous approach.

## One nit

`:771` sets the field through a `(entity as Entity & {attractionType?: string})` cast. `Entity.attractionType` is typed and `SHOW` is a real `AttractionTypeEnum` member, so the cast is not needed and it removes the only check on a field nothing else validates. This is the same argument that took the `as any` casts out of #294.

Minor consistency point while you are in there: `ENTITY_TYPE_OVERRIDES` is a keyed map at `:56`, but the `attractionType` assignment is a hardcoded `poi.id === 'P1G103'` at `:771`. A second entry would want the same map shape.

## Merge note, and it matters

**This PR and #296 auto-merge with no conflict in `disneylandparis.ts`, and the result silently breaks this PR.** Only `visibility.test.ts` conflicts. In the merged file #296's published-id filter runs on the raw `entity.id` one line before the alias:

```
1156:  if (publishedIds && !publishedIds.has(entity.id)) continue;   // #296
1160:  const scheduleId = ID_ALIASES[entity.id] ?? entity.id;        // this PR
```

`P1DA13` is hidden and never in `publishedIds`, so the 60 folded rows get dropped before the alias can move them. Fix is one line reordered:

```ts
const scheduleId = ID_ALIASES[entity.id] ?? entity.id;
if (publishedIds && !publishedIds.has(scheduleId)) continue;
```

Land this one first and #296 second, so the rebase has your `attaches the twin's schedule rows to P1G103` test going red if it gets the order wrong.

PR #298, opened 2026-08-13
fix(dlp): read park hours from the schedule feed, not the ride estate

cubehouse:
## Summary

Six walkthrough attractions (Discovery Arcade, Liberty Arcade, Sleeping Beauty Castle, La Galerie de la Belle au Bois Dormant, Horse-Drawn Streetcars, World Premiere) never appear in the wait feed, so `buildLiveData` derives their status from park hours. Those hours were the earliest start and latest end across queue-bearing *rides'* POI schedules.

Each park already publishes its own hours in the schedule feed this method fetches anyway. This reads them from there.

| | before | after |
|---|---|---|
| source | min/max across queue-bearing rides | each park's own schedule row |
| scope | one window shared by both parks | per park (`P1`, `P2`) |
| Extra Magic Hours | excluded | included |
| window crossing midnight | never inside it | handled |
| after park-local midnight | no data, everything CLOSED | correct |
| no hours available | emits nothing | emits `CLOSED` |
| extra HTTP requests | – | none, reuses the schedule fetch |

Entities, schedules and every other live-data path are untouched.

## The three problems, graded honestly

**A single ride sets the window for both parks.** Measured against the live feed today (2026-08-14):

```
published hours        P1 08:30 -> 22:00     P2 08:30 -> 22:00
derived from rides        08:19 -> 22:00  (both parks)
```

`08:19` is one attraction, *Ratatouille*, which sits in Walt Disney Studios (`P2`) — and its start time was setting the window for Disneyland Park too. Today that is an 11 minute error. The size is not bounded by anything: it is whatever the earliest or latest outlier happens to publish, and an earlier run measured `08:13`. The close side is the same shape in reverse, and became reachable when #292 made the railroad stations emittable, since they publish full-day windows on days they are flagged.

**One window for both parks.** `parkOpenStr` / `parkCloseStr` were module-scope, so `P1` inherited `P2`'s hours and vice versa. Today the two parks publish identical hours, so the only visible symptom is the cross-park leak above. On any day their hours diverge it becomes a straight wrong answer for one of them.

**It broke after park-local midnight.** The POI blob's `schedules` array only ever carries the date it was fetched — verified today, the entire blob holds exactly one distinct date, `2026-08-14` — and `getPOIData` is cached 12h. Once the park clock rolled past midnight, `todayStr` moved on, no POI row matched, `parkOpenStr`/`parkCloseStr` went null and every walkthrough fell to `CLOSED` until the cache expired. Observed live at 00:07 Paris. The schedule feed is fetched per date, so reading hours from there removes the dependency entirely.

## Extra Magic Hours counts

Each park publishes two rows:

```
P1  EXTRA_MAGIC_HOURS  08:30-09:30
P1  OPERATING          09:30-22:00
```

Both are used, unioned into one window. Extra Magic Hours is a hotel-guest preview rather than general opening, but guests are physically in the park and the rides publish windows covering it. Excluding it would report Big Thunder Mountain operating at 08:45 while Sleeping Beauty Castle, fifty metres away, read closed. Taking the union rather than picking one row also survives a day split by a private event, where `.find()` would have kept whichever half it reached first.

## Emitting CLOSED rather than nothing

The earlier revision of this PR returned `undefined` when no hours were available and pushed no row. That is wrong for the same reason #301 was: the wiki keeps a live-data row until something replaces it, so emitting nothing leaves a walkthrough last seen `OPERATING` reading `OPERATING` for as long as that lasts. Silence is not neutral.

`CLOSED` is also the right guess rather than a shrug: the query asks only for `OPERATING` and `EXTRA_MAGIC_HOURS` rows, so a park closed all day cannot produce a row at all. Absence is exactly what a genuine closure looks like.

## `shiftDateString`

A window closing after midnight (`09:30-01:00`) read as closing before it opened, so nothing was ever inside it. Fixing that needs a day step, and the obvious form is wrong:

```ts
new Date(`${todayStr}T00:00:00`)   // parses in the HOST's zone
```

On any host east of the park that silently fails to advance the date. Extracted to `src/datetime.ts` as `shiftDateString`, anchored at noon UTC where no offset or DST shift can cross a date boundary, and covered by a test that pins `process.env.TZ` across six zones from Kiritimati to Los Angeles. `buildSchedules` in this file, plus `futuroscope.ts` and `chimelong.ts`, still carry the naive form — out of scope here, worth a follow-up.

## Test plan

- [x] `npm test` — 77 files, 1730 tests, all passing
- [x] New `src/parks/dlp/__tests__/parkHours.test.ts`, 16 tests. Verified against pre-fix source: **7 fail**, including both regression cases (`ignores a queue-bearing ride that runs past park close`, `ignores a queue-bearing ride that opens before the park`). The first draft of those two passed against pre-fix code and were rewritten until they didn't.
- [x] Every test carries a canary walkthrough (`P1MA03`) with its own ±5 minute window asserted `OPERATING`, so a change that made the pipeline return nothing fails instead of passing vacuously. An earlier draft had exactly that hole.
- [x] 6 new tests for `shiftDateString` in `src/__tests__/datetime.test.ts`, including the TZ sweep. Mutating the helper back to the naive `T00:00:00` anchor fails the suite; before the sweep existed, that mutant survived.
- [x] Covered: inside/outside window, per-park isolation in both directions, a ride running past close, a ride opening early, no published hours, a `closed: true` park row, EXTRA_MAGIC_HOURS unioned, a window crossing midnight, and a schedule-feed failure
- [x] Live feed at 20:20 Paris (hours 09:30-22:00): all 6 walkthroughs `OPERATING`, La Galerie correctly `CLOSED` on its own schedule, 69 live rows, 131 entities, no row missing a status

## Merge note

Touches `buildLiveData` around the walkthrough block, so expect a small conflict with #294, which edits `isWithinWindow` in the same region. No overlap with #293, #295 or #296.

### DIFF
--- a/src/__tests__/datetime.test.ts
+++ b/src/__tests__/datetime.test.ts
@@ -14,6 +14,7 @@ import {
   localFromFakeUtc,
   hostnameFromUrl,
   constructDateTime,
+  shiftDateString,
 } from '../datetime.js';
 
 describe('DateTime Utilities', () => {
@@ -795,5 +796,56 @@ describe('DateTime Utilities', () => {
     test('handles http URL', () => {
       expect(hostnameFromUrl('http://localhost/api')).toBe('localhost');
     });
+  });
+});
+
+describe('shiftDateString', () => {
+  it('steps forward and back a day', () => {
+    expect(shiftDateString('2026-08-13', 1)).toBe('2026-08-14');
+    expect(shiftDateString('2026-08-13', -1)).toBe('2026-08-12');
+    expect(shiftDateString('2026-08-13', 0)).toBe('2026-08-13');
+  });
+
+  it('crosses month, year and leap-day boundaries', () => {
+    expect(shiftDateString('2026-08-31', 1)).toBe('2026-09-01');
+    expect(shiftDateString('2026-12-31', 1)).toBe('2027-01-01');
+    expect(shiftDateString('2027-01-01', -1)).toBe('2026-12-31');
+    expect(shiftDateString('2028-02-28', 1)).toBe('2028-02-29');
+    expect(shiftDateString('2028-03-01', -1)).toBe('2028-02-29');
+  });
+
+  it('crosses a DST boundary without slipping', () => {
+    // Europe/Paris springs forward on 2026-03-29 and back on 2026-10-25.
+    expect(shiftDateString('2026-03-28', 1)).toBe('2026-03-29');
+    expect(shiftDateString('2026-03-29', 1)).toBe('2026-03-30');
+    expect(shiftDateString('2026-10-24', 1)).toBe('2026-10-25');
+    expect(shiftDateString('2026-10-25', 1)).toBe('2026-10-26');
+  });
+
+  it('steps by more than one day', () => {
+    expect(shiftDateString('2026-08-13', 60)).toBe('2026-10-12');
+  });
+
+  it('throws on an unparseable date rather than returning a wrong one', () => {
+    expect(() => shiftDateString('not-a-date', 1)).toThrow(RangeError);
+  });
+
+  it('steps identically whatever timezone the host runs in', () => {
+    // The whole point of the helper. `new Date('2026-08-13T00:00:00')` parses
+    // in the host zone, so the naive form fails to advance on any host east of
+    // UTC — and CI runs UTC, where the naive form looks fine. Pinning the zone
+    // here is what makes that regression catchable.
+    const original = process.env.TZ;
+    try {
+      for (const tz of ['UTC', 'Pacific/Kiritimati', 'Pacific/Auckland', 'Asia/Tokyo', 'Europe/Moscow', 'America/Los_Angeles']) {
+        process.env.TZ = tz;
+        expect(shiftDateString('2026-08-13', 1), tz).toBe('2026-08-14');
+        expect(shiftDateString('2026-12-31', 1), tz).toBe('2027-01-01');
+        expect(shiftDateString('2026-08-13', -1), tz).toBe('2026-08-12');
+      }
+    } finally {
+      if (original === undefined) delete process.env.TZ;
+      else process.env.TZ = original;
+    }
   });
 });
--- a/src/datetime.ts
+++ b/src/datetime.ts
@@ -205,6 +205,27 @@ export function addDays(date: Date, days: number): Date {
 }
 
 /**
+ * Step a calendar date string by whole days, independent of the host timezone.
+ *
+ * `new Date('2026-08-13T00:00:00')` parses in the *host's* zone, so adding a
+ * day and reformatting in the park's zone silently returns the same date on
+ * any host east of that park. This anchors at noon UTC instead, which no
+ * offset or DST shift can push across a date boundary.
+ *
+ * @param dateStr Date in YYYY-MM-DD format
+ * @param days Whole days to add; may be negative
+ * @returns The resulting date in YYYY-MM-DD format
+ */
+export function shiftDateString(dateStr: string, days: number): string {
+  const anchored = new Date(`${dateStr}T12:00:00Z`);
+  if (!Number.isFinite(anchored.getTime())) {
+    throw new RangeError(`shiftDateString: unparseable date "${dateStr}"`);
+  }
+  anchored.setUTCDate(anchored.getUTCDate() + days);
+  return anchored.toISOString().slice(0, 10);
+}
+
+/**
  * Construct a full ISO 8601 datetime string from separate date, time, and timezone.
  * Replaces the per-park getXxxOffset() helpers that every park reimplemented.
  *
--- a/src/parks/dlp/__tests__/parkHours.test.ts
+++ b/src/parks/dlp/__tests__/parkHours.test.ts
@@ -38,6 +38,8 @@ function stubbedPark(opts: {
   walkthroughPark?: string;
   extraAttractions?: any[];
   waitTimes?: any[];
+  /** Per-entity rows in the schedule feed. Defaults to just the canary. */
+  entitySchedules?: any[];
 }): DisneylandParis {
   const park = new DisneylandParis();
   const today = parkToday();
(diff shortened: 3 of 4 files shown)


cubehouse:
Updated after review. Three things changed.

**The degraded state was a confident falsehood, and I had it backwards.** Emitting `CLOSED` when neither the walkthrough nor its park publishes hours asserts a closure nobody reported. Worse, the try/catch I added covered only the throwing path: `getScheduleForDate` returns `[]` on a bad 200 and `CacheLib` pins that for 24h, and past the publication horizon the feed genuinely returns activities with no P1/P2 rows at all. Neither goes through the catch. That is the same shape I objected to in #296, and it contradicted the rule I endorsed on #294 in this same file, that a venue with no published hours is skipped rather than reported closed. Walkthroughs now do the same: no row, so the last good value stands.

**A park day closing after midnight reported every walkthrough CLOSED all day.** `09:30-01:00` read as closing before it opened, so nothing was ever inside the window. `buildSchedules` already handles this shape.

**The tests were flaky on a real clock**, red between 22:30 and 01:30 Paris when a +90 minute window wraps past midnight. Gate tests may never be flaky, so the clock is now pinned with fake timers.

One extra thing came out of fixing that last one. Stepping to the next day via `new Date(\`${dateStr}T00:00:00\`)` parses in the **host's** timezone, so on a host at UTC+10 or beyond it lands on the previous day and the step silently does nothing:

```
TZ=Europe/Paris     new Date("2026-08-13T00:00:00") -> next="08/14/2026"
TZ=Australia/Sydney new Date("2026-08-13T00:00:00") -> next="08/13/2026"   <- no step
```

The suite caught it: 15/15 on Paris, UTC and Los Angeles, 14/15 on Sydney and Kiritimati. Stepping at noon UTC instead fixes it, and the suite now passes from Kiritimati to Kolkata.

**`buildSchedules` has the identical idiom** at the midnight-crossing branch, so its close times are wrong on eastern hosts too. Untouched here to keep this PR to one thing. It is latent for anything running UTC, real for a contributor in Australia or NZ. Happy to fix it separately.

Also pinned as a known limitation rather than left as a surprise: in the small hours we judge against the new day's window, because `buildLiveData` only fetches `todayStr`. Resolving that needs yesterday's row too.

Verified: full suite 1708 passing, `tsc --noEmit` clean, 5 of 15 tests fail against pre-fix source, harness 4/4 with 131 entities and 69 live rows, and the six walkthroughs resolving correctly against live hours.

cubehouse:
Updated after review. Two blocking findings, both correct.

**Omitting the row was the wrong answer to "we cannot see the hours", and I had just made the same mistake on #300.** The wiki keeps a live-data row until something replaces it, so emitting nothing leaves a walkthrough last seen OPERATING reading OPERATING indefinitely — the previous code at least wrote CLOSED. Worse, `fetchScheduleForDate` requests only `status: ['OPERATING','EXTRA_MAGIC_HOURS']`, so a park shut all day **cannot** produce a row: absence is exactly what a genuine closure looks like, which is precisely the case where CLOSED is most likely right. Measured, an empty or failed payload dropped 5 of 6 walkthroughs, and `getScheduleForDate` is cached 24h with no `validateResponse`, so a single GraphQL 200-with-errors stranded them for a full day. Now emits CLOSED, matching `fantawild.ts:689-698` and the nigloland precedent.

**EXTRA_MAGIC_HOURS now counts.** The reviewer measured that old and new disagree for exactly 60 minutes a day and that all 60 are the EMH window — the genuinely wrong part of the old open was only 11 minutes (08:19 vs 08:30). Excluding EMH meant reporting Big Thunder Mountain operating at 08:45 while Sleeping Beauty Castle fifty metres away read closed. The window is now the union of every published window for the day, which also covers a day split by a private event instead of silently taking the first row.

**Corrections to my own PR description**, which overclaimed:
- The old max-end never actually overshot: across 79 published days the latest ride end equals the published close.
- P1 and P2 publish identical windows on every one of those days, so the "conflated into one window" defect has no current effect.
- The real, measured defect was the open time: 08:19 against a published 09:30, of which 11 minutes were genuinely wrong.

The structural argument for reading each park's own row still holds; the stated rationale did not.

**`buildSchedules` is confirmed broken and left alone.** Its midnight-crossing branch uses the same `new Date(\`${dateStr}T00:00:00\`)` form, which fails on **any host east of Paris**, not UTC+10 as my comment said — verified producing `closingTime` before `openingTime` under Moscow, Tokyo and Auckland. Latent while the collector runs UTC. `futuroscope.ts` and `chimelong.ts` share the pattern. Worth its own change.

My day-step moves into `datetime.shiftDateString()`. The inline version was correct but **untestable in CI**: the naive form it guards against only misbehaves east of the park, and CI runs UTC, so the mutation survived. The helper's tests pin the zone explicitly — mutating it back to the naive anchor now fails under `TZ=UTC`.

**On the tests:** my "4 of 15 fail pre-fix" was wrong, it was 10 of 15, so the suite was stronger than I claimed. But 4 tests asserted only absence and would have gone green on a dead pipeline. Each now carries a canary walkthrough that resolves from its own schedule and must read OPERATING in the same call, so CLOSED is a decision rather than an empty result.

Verified: full suite 1715 passing, `tsc` clean, 7 of 16 fail against pre-fix source, green under six host timezones, harness 4/4.

cubehouse:
Verified against the live feed on 2026-08-14 at 18:30 Paris, both parks open on published hours 09:30 to 22:00. The park-hours rework is correct and does what the description claims.

```
P1  08:30-09:30 EXTRA_MAGIC_HOURS   09:30-22:00 OPERATING
P2  08:30-09:30 EXTRA_MAGIC_HOURS   09:30-22:00 OPERATING
```

Union gives 08:30 to 22:00 per park, exactly as described, and both parks are represented rather than sharing one window.

All six walkthroughs resolve, none from the ride estate:

```
P1MA00 Discovery Arcade                        OPERATING  P1
P1MA02 Horse-Drawn Streetcars                  OPERATING  P1
P1MA03 Liberty Arcade                          OPERATING  P1
P1NA04 Sleeping Beauty Castle                  OPERATING  P1
P1NA06 La Galerie de la Belle au Bois Dormant  OPERATING  P1
P2FD03 World Premiere                          OPERATING  P2
```

131 entities, 67 live rows, 0 rows missing a status.

`shiftDateString` is the right extraction and the noon-UTC anchor is the right anchor. Throwing a `RangeError` on an unparseable date rather than returning a silently wrong string is the correct call for a helper with 169 sibling call sites.

## One gap

`deriveWalkthroughStatus` at `:1076` still reads `poi.schedules` for an entity's own hours:

```ts
const own = (poi.schedules || []).find((s) => s.date === todayStr);
```

That is the same POI-blob-carries-only-the-fetch-day staleness the description lists as problem 3, so only the park-hours half of it is fixed. Impact is small, because past local midnight both the own-hours path and the park-hours path read `CLOSED` on a 22:00 close. It is not nothing though: La Galerie is the one walkthrough with its own shorter window, and after rollover it silently falls back to the full park window until the 12h cache turns over. On a late-close night the difference becomes visible.

#294 hoists `todaySchedules` from `getScheduleForDate(todayStr)` in the same method for exactly this reason. Sourcing `own` from there removes the dependency completely and costs no extra request.

## Notes

- **Three `getScheduleForDate(todayStr)` call sites** once this merges with #294 (`:919` showtimes, `:1023` here, plus #294's dining). Cached at 24h so no extra HTTP, but collapse them onto one local.
- **`nowMs <= c` at `:1058` is inclusive**, while #294 moves the dining window to half-open `nowMs < endTime` to match `attractionsiov1.ts:303-309`. Two windows in one file disagreeing by a minute. Worth settling when they merge.
- `close < open` at `:1055` compares the raw strings. Safe here since both sides come from the same feed row and every live row carries seconds, but #296 hit the mixed `HH:MM` / `HH:MM:SS` version of this and normalised. Same treatment would make it robust rather than incidentally correct.
- This branch conflicts with none of #293 through #297, so it can land at any point in that sequence.

PR #299, opened 2026-08-13
feat(harness): report live data and schedules keyed to unpublished entities

cubehouse:
## Summary

Live data and schedules are keyed by entity id. A row keyed to an id that `getEntities()` never emits cannot be looked up by any consumer. It usually means a `build*` path is reading a wider upstream feed than `buildEntityList` does. The harness validated schedule *structure* but never checked that the ids resolve, so this class of defect was invisible.

This adds the cross-check for both `getLiveData()` and `getSchedules()`, printed per park and rolled up in the run summary.

```
   ✓ Found 133 schedule(s) with 14139 total days (36435ms)
     ⚠ Schedules for unpublished entities: 36 (P1MG86, H03R00, H02R05, D01R02, H07R00, +31 more)
       (5849 of 14139 days affected)
```

```
Rows keyed to unpublished entities (2 park(s)):
  ⚠ Busch Gardens Tampa                      12 live rows
  ⚠ Busch Gardens Williamsburg               7 live rows
```

## What it found: 10 of 80 parks

All 80 park ids run through `testPark()`. **10 carry orphans**, measured in a single sweep on 2026-08-14 at 10:49Z:

| park | live orphans | schedule orphans |
|---|---|---|
| sixflags | 42 of 1457 rows | 0 |
| disneylandparis | 0 | 36 of 133 rows (5,849 of 14,139 days) |
| seaworldsandiego | 15 of 87 | 0 |
| buschgardenstampa | 12 of 71 | 0 |
| universalstudios | 11 of 48 | 0 |
| seaworldorlando | 10 of 76 | 0 |
| seaworldsanantonio | 9 of 43 | 0 |
| buschgardenswilliamsburg | 7 of 65 | 0 |
| phantasialand | 7 of 58 | 0 |
| sesameplacephiladelphia | 3 of 117 | 0 |

Orphaned **live data** is the common case, 9 parks to 1. Counts drift with the upstream feeds — sixflags moved 1447 → 1457 rows between two runs the same week, and USH moved 11 → 10 within four hours of the same day. Ten is also a *lower bound*: a park whose `getEntities()` fails reports no orphans by design, so any park failing that run is silent here.

## What those orphans actually are

Each park below was re-sampled during its own operating hours, because a park sampled overnight tells you nothing about whether its orphans carry data.

- **Six Flags** (42) — the union of `/venue-status` and `/wait-times` is deliberate, and the comment at `sixflags.ts:956-984` explains why: the vendor sometimes publishes a live wait for a ride it has dropped from that roster, and enumerating the roster alone discards it. The incident recorded there is the *pre-union* behaviour — at Canada's Wonderland on 2026-07-23 The Daredeviler's 60 minute wait was thrown away and the wiki kept serving a CLOSED record frozen since April. The union is what stops that, and its side effect is that a ride absent from `/poi` too now surfaces here as an orphan **carrying a real wait**. So this park is structurally capable of orphans-with-data, by design.
  Sampled 2026-08-14 14:35Z with the parks open: none of the 42 carried data. 41 are shows publishing a bare `OPERATING`, and the 42nd (`RIDE-001-00181`) is `CLOSED` with an empty `queue.STANDBY: {}`. Scanning the operating-hours show lists forward through September, those 41 show ids carry times on **no** date, and no ride id is currently absent from both feeds. The right fix here is publishing the missing entities, not suppressing the rows.
- **SeaWorld family** (15 / 12 / 10 / 9 / 7 / 3) — cosmetic. Every orphan is a ShowTimes row with zero times, publishing a bare `CLOSED`; there are no orphaned wait-time rows anywhere in the family. Sampled during operating hours (Orlando and Busch Gardens Tampa at 14:45Z, 10:45 local): zero orphans carrying data. Querying the availability feed forward, the orphan show ids carry times on no date at any of the six parks, while the resolved shows do.
- **Phantasialand** (7) — cosmetic, but only with #301 applied, which matters because this branch was cut before it. Sampled 14:40Z with #301 in the tree and the park open (16:40 local): all 7 bare `CLOSED`, none carrying a queue or showtimes. Without #301 the same 7 were the frozen relics it retires — two carried standby waits (Black Mamba 40 min, Colorado Adventure 5 min) and two carried showtimes. #301 is already on `main`, so the merged result is the sampled one.
- **Universal Studios Hollywood** (10-11) — **the one with real loss.** Sampled 14:47Z: 2 of the 10 carry live showtimes. `ush.mms.minion.dance.party` (6 showtimes) sits in a namespace with **no** entries in the places feed at all — an id-namespace mismatch, a straight coverage gap. `ush.upper_lot.shows.studio_tour_spanish` (6 showtimes) is different: it *is* in the places feed and is dropped on purpose by `isEventVariantAlias` as a language variant of the canonical Studio Tour. Whether losing its separate departure times is correct dedup or real loss is a judgement call, not a bug. The remaining orphans split the same way — some absent from `/places`, some deliberately excluded (CityWalk via `NON_SURFACED_VENUE_PARENT`, other Studio Tour variants).
- **Disneyland Paris** (36 schedule rows, 41% of its days) — dead weight: 14 character meet & greets, 21 hotel and Disney Village restaurants and bars, and Mickey's PhilharMagic, whose published SHOW entity already carries its own schedule. Addressed in #296.

So the honest summary: Universal Studios Hollywood is losing showtimes today; Six Flags produces orphans-with-data by design whenever the vendor drops a still-operating ride, though none of its current 42 carry any; DLP's are dead weight; the SeaWorld family's and Phantasialand's are cosmetic, verified with the parks open.

## Why a warning, not a failure

An orphan row is always wrong, so failing is defensible, and now that the count is known (10) it would not be an unbounded change. Two reasons to keep it a warning:

- **The harness is not in CI.** `.github/workflows/unit_test.js.yml` runs `npm ci`, `npm run build`, `npm test`; `docs.yml` runs `npm run docs`. Neither invokes `npm run dev`, so a hard failure buys no automated signal.
- **Most of the ten are cosmetic**, per the breakdown above, and the two that matter need per-park investigation rather than a mechanical fix.

Worth promoting to a failure once those ten are cleared. One caveat for whoever does that: entity data is cached far longer than live data in several parks (`seaworld.ts:188` 12h vs `:197` 60s), so on a warm cache an upstream addition can appear in live data before the entity list learns about it, producing a transient orphan that is not a code defect. Pair the promotion with a cache clear, or accept the flake.

This follows the precedent already in the file: structural violations throw, things that are legitimately variable are counted and printed with `⚠`.

## What this check does not catch

Worth stating plainly, because I overestimated it at first. A park that filters its live rows against the entity list passes this check regardless of whether the data landed on the *right* entity. It measures resolvability, not correctness. It would have gone green on a change that rewrote ids onto plausible-looking but wrong entities.

That caveat is not hypothetical: **20 of the 80 destinations can never report an orphan**, because their shared base classes intersect live rows with the entity id set before emitting (`te2.ts:775-783`, 4 destinations; `attractionsiov1.ts:1154-1164`, 16). They pass by construction.

Its value is narrower than "finds bugs": it surfaces parks whose two build paths disagree about what exists. That is worth a look, and at Phantasialand it paid off directly — all 7 of its orphans turned out to be stale signage rows the feed never prunes, published as current, fixed in #301.

## Implementation

`findOrphanIds(rows, publishedIds)` is exported as a pure function so it can be unit tested. It returns nothing when the published set is empty, so a failed `getEntities()` reports no orphans rather than flagging every row and burying the real failure.

Cost is not measurable: 2.45 ms total across all 80 parks, max 0.33 ms on sixflags (1457 rows against 2558 entities, in a 152 s park run). An isolated benchmark on the same shape puts it at 0.068 ms warm, 0.474 ms cold.

## Test plan

- [x] `npm test` — 75 files, 1700 tests, all passing
- [x] `src/__tests__/orphanEntityCheck.test.ts` — mutation-tested: 10 mutants of `findOrphanIds` (dropped guard, `row.id` for `row?.id`, array for Set, case-insensitive compare, prefix match, counting `''`, report-everything, inverted membership, always-empty, first-only) — all 10 killed
- [x] `src/__tests__/orphanReporting.test.ts` — drives the real `testPark()` against a stub Destination; 7 reporting mutants killed
- [x] `npm run dev -- disneylandparis` — reports 36 ids and the day count, matching an independent count
- [x] `npm run dev -- --category "Busch Gardens"` non-verbose — roll-up appears in the summary, which is the mode the first revision missed
- [x] Verified silent when `getEntities()` fails: walibibelgium fails on an id-charset error and reports no orphans despite 15 live rows
- [x] Severity of every entry above re-sampled during that park's operating hours, not overnight

### DIFF
--- a/src/__tests__/orphanEntityCheck.test.ts
+++ b/src/__tests__/orphanEntityCheck.test.ts
@@ -1,0 +1,50 @@
+import {describe, it, expect} from 'vitest';
+import {findOrphanIds} from '../testRunner.js';
+
+/**
+ * Live data and schedules are keyed by entity id. A row keyed to an id that
+ * getEntities() never emits cannot be looked up by any consumer, so the
+ * harness reports it. Disneyland Paris was publishing 36 such schedule ids,
+ * 42% of its schedule days, because buildSchedules read a wider upstream feed
+ * than buildEntityList did.
+ */
+describe('findOrphanIds', () => {
+  const published = new Set(['P1', 'P2', 'P1RA00']);
+
+  it('returns nothing when every row is published', () => {
+    expect(findOrphanIds([{id: 'P1'}, {id: 'P1RA00'}], published)).toEqual([]);
+  });
+
+  it('finds a row keyed to an unpublished id', () => {
+    expect(findOrphanIds([{id: 'P1'}, {id: 'H03R00'}], published)).toEqual(['H03R00']);
+  });
+
+  it('reports each orphan id once however many rows carry it', () => {
+    const rows = [{id: 'H03R00'}, {id: 'H03R00'}, {id: 'H03R00'}, {id: 'D01R02'}];
+    expect(findOrphanIds(rows, published)).toEqual(['H03R00', 'D01R02']);
+  });
+
+  it('stays quiet when the entity list is empty', () => {
+    // getEntities() failed or was skipped. Reporting every row as an orphan
+    // would bury the real failure under noise.
+    expect(findOrphanIds([{id: 'H03R00'}, {id: 'D01R02'}], new Set())).toEqual([]);
+  });
+
+  it('ignores rows with no id', () => {
+    expect(findOrphanIds([{}, {id: undefined}, {id: ''}], published)).toEqual([]);
+  });
+
+  it('survives null and undefined rows', () => {
+    const rows = [null, undefined, {id: 'H03R00'}] as unknown as Array<{id?: string}>;
+    expect(findOrphanIds(rows, published)).toEqual(['H03R00']);
+  });
+
+  it('returns an empty list for an empty input', () => {
+    expect(findOrphanIds([], published)).toEqual([]);
+  });
+
+  it('is exact about ids rather than matching loosely', () => {
+    // 'P1' is published; 'P10' and 'p1' are not.
+    expect(findOrphanIds([{id: 'P10'}, {id: 'p1'}], published)).toEqual(['P10', 'p1']);
+  });
+});
--- a/src/__tests__/orphanEntityCheck.test.ts
+++ b/src/__tests__/orphanEntityCheck.test.ts
@@ -1,27 +1,30 @@
 import {describe, it, expect} from 'vitest';
-import {findOrphanIds} from '../testRunner.js';
+import {findOrphanIds} from '../orphanCheck.js';
 
 /**
  * Live data and schedules are keyed by entity id. A row keyed to an id that
  * getEntities() never emits cannot be looked up by any consumer, so the
  * harness reports it. Disneyland Paris was publishing 36 such schedule ids,
- * 42% of its schedule days, because buildSchedules read a wider upstream feed
- * than buildEntityList did.
+ * covering most of its schedule days, because buildSchedules read a wider
+ * upstream feed than buildEntityList did.
  */
 describe('findOrphanIds', () => {
-  const published = new Set(['P1', 'P2', 'P1RA00']);
+  // Built per test. Shared across tests it silently masked whether the
+  // implementation mutates the caller's set: a mutating version "failed" only
+  // because it poisoned later tests, which would evaporate on a reorder.
+  const mkPublished = () => new Set(['P1', 'P2', 'P1RA00']);
 
   it('returns nothing when every row is published', () => {
-    expect(findOrphanIds([{id: 'P1'}, {id: 'P1RA00'}], published)).toEqual([]);
+    expect(findOrphanIds([{id: 'P1'}, {id: 'P1RA00'}], mkPublished())).toEqual([]);
   });
 
   it('finds a row keyed to an unpublished id', () => {
-    expect(findOrphanIds([{id: 'P1'}, {id: 'H03R00'}], published)).toEqual(['H03R00']);
+    expect(findOrphanIds([{id: 'P1'}, {id: 'H03R00'}], mkPublished())).toEqual(['H03R00']);
   });
 
   it('reports each orphan id once however many rows carry it', () => {
     const rows = [{id: 'H03R00'}, {id: 'H03R00'}, {id: 'H03R00'}, {id: 'D01R02'}];
-    expect(findOrphanIds(rows, published).sort()).toEqual(['D01R02', 'H03R00']);
+    expect(findOrphanIds(rows, mkPublished()).sort()).toEqual(['D01R02', 'H03R00']);
   });
 
   it('stays quiet when the entity list is empty', () => {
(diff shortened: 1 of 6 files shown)


cubehouse:
Updated after review, plus a correction to the description that was mine to make.

**The Phantasialand claim in the original description was false and I have rewritten it.** It said the park was "actively broken" and publishing no wait times at all. I had measured at night with the park shut, so of course no waits were present, and the rows I called broken were frozen relic rows. Re-measured on current `main`: 7 orphans, **none carrying a queue or showtimes**, and 29 published rows with wait times. Phantasialand belongs in the same harmless category as Six Flags and the SeaWorld parks. The description now carries a severity breakdown instead of a count: one park losing real data (Universal Studios Hollywood), one with a documented latent risk (Six Flags), the rest cosmetic.

I also added a section on what this check does **not** catch, since I overestimated it. It measures resolvability, not correctness — a park that filters its live rows against the entity list passes regardless of whether the data landed on the right entity. It would have gone green on #300's regression.

**Two test findings fixed:**
- The empty-set guard could be weakened to `size <= 1` with everything still green: no test used a one-entity list, so a destination publishing exactly one entity would have reported no orphans. Pinned.
- The shared module-scope fixture meant "does not mutate the caller's set" was never actually asserted — a mutating implementation only failed by poisoning later tests, which a reorder would have hidden. Fixture is now per-test and the property has its own assertion. Verified both mutants now die deliberately.

**The reporting code had no tests.** Added seven driving the real `testPark`: the five-id truncation, the `+N` tail, exactly five and exactly six, a single orphan, the day sum over rows rather than ids, and silence when the entity list cannot be built. All seven kill at least one of seven mutants I threw at the truncation and day-count arithmetic.

**Moved `findOrphanIds` to `src/orphanCheck.ts` and exported it.** You were right that `testRunner` is the wrong home: it is absent from `src/index.ts` and `package.json` declares no subpath export, so the collector cannot reach the check even though it ships — and the invariant matters on the push path, where orphan rows reach consumers continuously, rather than when a human runs a park. It also stops a ten-line pure function's unit test from booting the SQLite cache and the HTTP queue's interval timer.

Also fixed: the summary roll-up assigned rather than accumulated per test name, and printed "12 live" where it meant rows.

Verified: 76 files, 1709 tests passing, `tsc` clean, and the non-verbose category run still reports `⚠ Busch Gardens Tampa 12 live rows`.

cubehouse:
No blocking findings. The check is well scoped and the honesty about what it does not catch is the most useful part of the description.

`tsc --noEmit` clean, suite green, no conflicts against any of #293 through #298 or #302.

What I checked rather than assumed:

- **The empty-set guard is load-bearing and correct.** `findOrphanIds` returning `[]` when `publishedIds.size === 0` is what stops a failed `getEntities()` reporting every row as an orphan and burying the real failure. `testPark` feeds it an empty set only on that path, so the two halves agree.
- **The `null`/`undefined` row tolerance is real.** `row?.id` rather than `row.id` matters because this runs over whatever a park's `build*` returned, before any sanitisation.
- **It does not mutate the caller's set**, and there is a test pinning that, which is the sort of thing that only ever gets caught once.
- **Independent confirmation on DLP.** Same minute, same feed: `main` gives 135 schedule rows with 38 orphans. #297 takes that to 134 / 37 by folding the PhilharMagic twin, and #296 takes it to 97 / 0. So the check tracks the two PRs that are meant to move it, in the right direction, by the right amount.

## Notes

- `src/orphanCheck.ts` is exported from `src/index.ts`, so `findOrphanIds` becomes public API. That reads deliberate given the stated intent of asserting the same invariant on the collector's push path, just worth being explicit that it is now a compatibility surface.
- The summary roll-up in `test.ts` counts distinct orphan ids and prints them as "N live rows" / "N schedule rows". Those numbers coincide today, because both `getLiveData()` and `getSchedules()` emit at most one row per id. Fine as it stands; the wording would drift if that ever stopped being true.
- Agreed on keeping it a warning rather than a failure while the harness is outside CI. The caching caveat you flag at the end is the real obstacle to promoting it: entity data outliving live data produces a transient orphan that is not a defect, and that would need a cache clear paired with the promotion.

PR #300, opened 2026-08-13
fix(phantasialand): attach wait times to the entities that exist

cubehouse:
## Summary

Phantasialand publishes **no wait times at all**, and its two headline coasters read closed while they are running. The data is fetched correctly on every poll and then filed under ids no consumer can resolve.

| measured against the live feed | before | after |
|---|---|---|
| entities | 110 | 110 |
| live rows | 58 | 51 |
| rows keyed to an unpublished id | **7** | **0** |
| **entities carrying a queue** | **0** | **2** |
| entities carrying showtimes | 13 | 13 |

## What was happening

The POI feed carries several records per venue: one with `seasons`, which `buildEntityList` publishes, and retired or placeholder copies with `seasons: []`, which it skips at `phantasialand.ts:385`. `buildLiveData` at `:458` emitted a row for every signage `poiId` and never consulted the entity list.

The signage feed reports against whichever copy it likes:

| venue | published entity | its live row | unpublished twin | its live row |
|---|---|---|---|---|
| Black Mamba | `52` (seasons=2) | `{"status":"CLOSED"}` | `225` (seasons=0) | `{"status":"OPERATING","queue":{"STANDBY":{"waitTime":40}}}` |
| Colorado Adventure | `224` (seasons=2) | `{"status":"CLOSED"}` | `25` (seasons=0) | `{"status":"OPERATING","queue":{"STANDBY":{"waitTime":5}}}` |

So an inverted coaster and the mine train both showed CLOSED while queueing at 40 and 5 minutes, and every other wait the park produced was discarded downstream.

## The fix

Live rows are aliased onto the published id, matching on `title`, which is exact across the copies. Guardrails:

- **A title is only aliased when exactly one of its records is published.** `Focaccia Mexico` genuinely has two published records and `Nobis` has three, none published; both stay unmapped rather than guessed at. Picking one would put a wait time on the wrong venue.
- **Rows that still resolve to nothing are dropped**, rather than published against a dead id. That covers services and photo points (`ATM`, `Face painter`, `Lockers`) which are not entities in any copy.
- **Where two signage entries land on the same entity** — a bare CLOSED for the published copy and the real state for the other — the more informative row wins, scored on queue, showtimes and status. Order-independent, and tested both ways round.

The publish rule moves into a single `publishedEntityType()` that both paths call. Two copies of that rule drifting apart is the actual root cause, so leaving them separate would invite the same bug back.

## How it surfaced

The orphan check in #299. Nobody was looking at Phantasialand; it passed the harness 4/4 before and after, because a row keyed to a non-existent entity was not something anything checked.

## Test plan

- [x] `npm test` — 75 files, 1702 tests, all passing
- [x] New `liveIdAliases.test.ts`, 9 tests, **7 fail against pre-fix source**
- [x] Covered: a wait landing on the published entity, the bare-CLOSED merge in both orderings, no duplicate ids in the output, showtimes aliasing as well as waits, a venue with no published copy, a copy unpublished only by category, a title with two published records left unguessed, an unambiguous id untouched, and the overall invariant that no row is emitted for an unpublished id
- [x] `npm run dev -- phantasialand` — 4/4, 110 entities, 51 live rows
- [x] Live feed: Black Mamba and Colorado Adventure now carry `OPERATING` with their real waits on entities `52` and `224`

### DIFF
--- a/src/parks/phantasialand/__tests__/liveIdAliases.test.ts
+++ b/src/parks/phantasialand/__tests__/liveIdAliases.test.ts
@@ -1,0 +1,190 @@
+import {describe, it, expect, vi, afterEach} from 'vitest';
+import {Phantasialand} from '../phantasialand.js';
+
+/**
+ * Phantasialand's POI feed carries several records per venue: one with
+ * seasons, which becomes the entity, and retired or placeholder copies with
+ * `seasons: []`, which do not. The signage feed reports live state against
+ * whichever copy it likes.
+ *
+ * Live: Black Mamba's 40-minute wait arrived under unpublished id 225 while
+ * published entity 52 got a bare CLOSED, and across the whole park not one
+ * published entity carried a queue. The live rows are aliased onto the
+ * published id so the wait reaches something a consumer can look up.
+ */
+const SEASON = [{from: '2026-01-01', to: '2026-12-31'}];
+
+function poi(over: Record<string, unknown>) {
+  return {
+    id: 1,
+    title: 'Ride',
+    category: 'ATTRACTIONS',
+    seasons: SEASON,
+    tags: [],
+    entrance: {world: {lat: 50.79, lng: 6.87}},
+    ...over,
+  };
+}
+
+function stubbedPark(pois: any[], signage: any[]): Phantasialand {
+  const park = new Phantasialand();
+  vi.spyOn(park as any, 'getPOI').mockResolvedValue(pois);
+  vi.spyOn(park as any, 'getSignage').mockResolvedValue(signage);
+  return park;
+}
+
+/** The published live row for an id, if one was emitted. */
+async function rowFor(park: Phantasialand, id: string) {
+  const live = await park.getLiveData();
+  return live.find((l) => l.id === id);
+}
+
+describe('Phantasialand live-id aliasing', () => {
+  afterEach(() => vi.restoreAllMocks());
+
+  it('lands a wait time reported against an unpublished copy on the published entity', async () => {
+    const park = stubbedPark(
+      [
+        poi({id: 52, title: 'Black Mamba'}),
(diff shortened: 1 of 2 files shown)


cubehouse:
Closing this. The premise is wrong.

I diagnosed it from a measurement taken while the park was shut. With the park closed every live row correctly reports `open: false, waitTime: null`, so "0 entities carrying a queue" was the right answer rather than a bug — and the only rows carrying queues were relics.

The unpublished twins are not the same venue reported under a different id. They are dead rows the signage feed never deletes. Live rows are regenerated every couple of minutes; retired ones stay frozen. Measured at 08:23 Berlin with the park open:

```
poiId 52  (Black Mamba, PUBLISHED)   wait=5    open=true   age=0 hours
poiId 225 (unpublished twin)         wait=40   open=true   age=10,077 hours
poiId 224 (Colorado Adv, PUBLISHED)  wait=null open=false  age=0 hours
poiId 25  (unpublished twin)         wait=5    open=true   age=23,558 hours
```

So this change would have replaced a correct 5-minute wait on Black Mamba with a 420-day-old 40, and reported Colorado Adventure as OPERATING off a 981-day-old row while it was genuinely shut. It reduces correct wait coverage from 12 entities to 11. `main` has been publishing these correctly all along.

All 38 fresh rows are already keyed to published ids, so the aliasing rescued nothing.

Two further problems, for the record: `livenessScore` resolved ties by feed position, and `getLiveIdMapping()` made a POI outage silently zero every live row for up to 6 hours.

There is a real bug here, just a different one — 13 stale rows land on published entities and are being served right now, including Mystic Winter Castle publishing nine showtimes in August off a 169-day-old observation. Replacement PR filters on row freshness, which needs no POI dependency and makes both the aliasing and the scoring unnecessary.

PR #302, opened 2026-08-14
chore: add containerised dev environment

BeyondVertical:
Picks up #271, which was closed after review. First commit is lgnap's original state unchanged (co-authored); everything after it works through the two review rounds one finding per commit, so the diff stays readable commit by commit.

A containerised dev environment: `make build && make park PARK=efteling` on a host with nothing but Docker installed. The container serves the API and the test harness; it does not build or serve the web UI.

Verified on Docker 29.7.2 / Compose v5.4.0, host uid 1000, on a checkout that has `core.fileMode=false` and is shared with Windows — both relevant below.

## What it adds

| File | |
|---|---|
| `Dockerfile` | `node:24-slim`, deps baked into a cached layer, uid-independent `HOME` and npm cache |
| `docker-compose.yml` | one service, bind-mounted tree, runs as the host uid/gid |
| `parksapi-entrypoint.sh` | creates `.env`, installs deps under a lock, verifies the install before trusting it |
| `Makefile` | every npm script as a target, per-checkout compose project |
| `.dockerignore` | default-deny allowlist |
| `.gitattributes` | LF pinning for the three new text files |
| `.github/workflows/dev_container.yml` | `paths:`-gated job covering all of the above |

## Second review round

**`.dockerignore` was root-anchored.** A working tree nested inside the checkout put its `cache.sqlite` — the `@cache` store, carrying live bearer tokens for nine destinations — into an image layer, along with a nested `.env` and untracked scratch files. It is default-deny now: `*`, then back in the seven paths the image needs. Reproduced with a planted nested checkout before fixing, and the built image is empty of all of them after.

**Every checkout was the same compose project**, so `make clean` in one tree tore down containers in another. `COMPOSE_PROJECT_NAME` is now derived from `CURDIR`. The earlier clean-slate run only looked isolated because the clone was named `parksapi-clonetest`; it is named `parksapi` now.

**`HOME=/home/node` broke the entrypoint's own advice.** Moving `HOME` and `npm_config_cache` into the image was not sufficient on its own: `npm cache clean` leaves `/tmp/.npm` behind owned by the build uid, so a mismatched uid hit the same EACCES one path along. Both that and node's compile cache are removed at the end of the install layer.

**Dependency installs are serialised** behind an `mkdir` lock with a double-checked stamp, `sha256sum` is no longer piped (so a missing lockfile fails under `set -e` instead of continuing with an empty hash), and a stamp is only trusted when `node_modules/.bin/tsx` actually exists — `npm ci` exits 0 on a failed optional dependency, so a good stamp could sit on a broken tree.

**CI** gained `package.json`, `package-lock.json` and `.gitattributes` to its `paths:` gate, plants files to prove the allowlist holds, and no longer passes `--retry=1` — that flag was broader than the problem it papered over. The underlying network dependency in the gate suite is #304.

**The web UI is out.** `src/webServer.ts` serves the SPA from `web-ui/dist`, which is both gitignored and dockerignored, so nothing in the container can produce it. The `package.json` change is reverted and this PR no longer touches that file at all. The 8080 → 8888 corrections in `WEB_ADMIN.md` stay: they fix a genuine pre-existing error against `src/webServer.ts`.

## First review round

Compose reading `./.env` (`--env-file /dev/null`), root-owned `node_modules` (named volume dropped, container runs as the host uid), stale deps (lockfile stamp), `make clean` leaving containers behind, `--category` quoting, CRLF handling, the uid error message, an unconditional write probe, and the CI job itself. Details in the first review comment.

Two things worth recording from that round: pre-creating `node_modules` in the image does not work — the engine creates the host mountpoint as `root:root` before image content or `USER` matter — and `NODE_PATH` is unusable because the repo is `"type": "module"`.

## Deliberate deviations

- **No `git` in the image.** The tree is bind-mounted, so git runs on the host; a README line says so. Keeps ~40 MB out.
- **`make build` keeps its name** (compose-idiomatic). The misleading README sentence is fixed instead.
- **Base tag stays `node:24-slim`** rather than pinning a patch version that would need bumping.
- **Podman is untested** — none on this machine. The documented recipes follow from the uid mapping but are not verified.

## Runs

Every target on the real tree: `list` (80 destinations), `test` (77 files, 1776 tests, no retry, matching the host suite exactly), `compile`, `park`, `category CATEGORY="Alton Towers"`, `health ARGS=efteling`, `cache-clear ARGS=efteling`, `shell`, `rebuild --no-cache`, `clean`, `clean-all`. `find . -uid 0` empty throughout.

Fresh clone, named `parksapi`, no `.env`, no `node_modules`:

```
make build && make park PARK=efteling   -> 4/4
.env            0 bytes, host-owned
node_modules    host-owned, stamp matches the lockfile
find -uid 0     empty
git status      clean
image           parksapi-4a1ec80d-parksapi
make clean-all  removed only its own image; the main tree's survived
```


lgnap:
Thanks ;-)

cubehouse (changes requested):
This is careful work and the point-by-point rework of the #271 review is exactly what I hoped for. The uid mapping, the dep stamp, the CRLF handling and the `--env-file /dev/null` flag all hold up. I reviewed it across three passes (image and entrypoint, compose and Makefile, CI and docs) and most of it survived.

Three things have to land before it merges. The first is the serious one.

## Blocking

### 1. `.dockerignore` is root-anchored, and a nested checkout puts credentials in an image layer

The file's own header says patterns need `**/` to catch nested files. 22 of the 26 patterns then omit it. Only `node_modules`, `dist`, `*.har`, `testout_*`, `temp*.json` and `temp*.js` are nested-safe. `.env`, `cache.sqlite*`, `.git`, `.claude/settings.local.json`, `datacache`, `snapshots` and the rest match at the root only.

That is not theoretical. My checkout has a working tree nested inside it, and building this PR's context produced a 283 MB context of which 264 MB was that nested tree's `cache.sqlite`, confirmed present in the built image while the root-level `cache.sqlite` was correctly excluded.

`cache.sqlite` is the `@cache` store. Nine park implementations put live bearer tokens through it:

```
attractionsio/attractionsiov1.ts:706   getInstallationToken
universal/universal.ts:581             getUdxToken
usj/universalstudiosjapan.ts:175       getToken
uss/universalsingapore.ts:296          getToken
europapark/europapark.ts:265           getToken
parcsreunidos/parcsreunidos.ts:265     getAccessToken
blackpool/blackpoolpleasurebeach.ts:77 getAccessToken
energylandia/energylandia.ts:602       getIdToken
flamingoland/flamingoland.ts:231       getIdToken
```

So a nested `cache.sqlite` in an image layer is live credentials in an image layer. A nested `.env` is the same problem in a more direct form; mine happens to be a symlink so only the dangling link copied, which is luck rather than design.

Minimum fix is `**/` on every pattern. The better fix is default-deny, because mirroring `.gitignore` gives no protection against untracked-but-not-ignored files, and an ad-hoc scratch file sitting in my working tree copied straight in. Something like `*`, then `!src`, `!package.json`, `!package-lock.json`, `!tsconfig.json` and the handful of others the build actually needs. That closes this whole class rather than the instances we happened to enumerate.

### 2. Every checkout is the same compose project, so `make clean` reaches across them

`docker-compose.yml` has no `name:` and the Makefile exports no `COMPOSE_PROJECT_NAME`, so compose derives the project from the directory basename. Two checkouts both named `parksapi` render identically:

```
.../a/parksapi -> project= parksapi
.../b/parksapi -> project= parksapi
```

Same project label, same network, same built image. So `make clean` in a review clone tears down a `make dev` running in the main checkout, and `make clean-all` with `--rmi local` deletes the image out from under it.

The clean-slate run in the description tested a clone named `parksapi-clonetest`, which is why it looked namespaced. Rename that clone to `parksapi` and the isolation goes away.

One line next to the UID/GID exports at `Makefile:19-20`:

```make
export COMPOSE_PROJECT_NAME := parksapi-$(shell printf '%s' '$(CURDIR)' | sha256sum | cut -c1-8)
```

### 3. `HOME=/home/node` makes the entrypoint's own advice fail

`Dockerfile:7` sets `HOME=/home/node`, owned by uid 1000, and only `docker-compose.yml:4` overrides it to `/tmp`. So the remediation `parksapi-entrypoint.sh:10` prints is broken for exactly the users who see it. Running the line as written:

```
docker run --user 1001:1001 -v <repo>:/app <image>
```

gives `EACCES` on `mkdir /home/node/.npm/_cacache`, aborts the install, and leaves a partial `node_modules` owned by 1001 in the bind mount. Since the image advertises a bare `docker run` path, the uid-independent `HOME` and `npm_config_cache` belong in the image or the entrypoint, not only in compose.

## Should fix

**Concurrent `npm ci` into the shared bind mount.** `parksapi-entrypoint.sh:29` installs into the host tree with no lock. `make test & make list` on a fresh clone has both entrypoints find no stamp, both start `npm ci`, and one `rm -rf node_modules` lands mid-unpack of the other. Whichever finishes last writes a valid stamp over the corrupted tree, so it stays broken until the lockfile changes. `make deps` fired during a running `make dev` is the same shape. An `flock` or `mkdir` lock around `ensure_deps` covers it.

**`sha256sum` failure cannot trip `set -e`.** `parksapi-entrypoint.sh:24` pipes into `cut`, so the pipeline status is `cut`'s and always 0. A missing lockfile continues with `want=""` and surfaces as 40 lines of `npm help ci`. Dropping the pipe and using `${want%% *}` fixes it, and also removes the case where an empty stamp file compares equal to an empty `want` and skips the install entirely.

**A stale stamp is trusted unconditionally.** `npm ci` exits 0 when an optional dependency fails to fetch, so one flaky registry moment on first start writes a good stamp over an incomplete tree, and every later start skips the repair. The symptom is `tsx: not found` with no hint that `PARKSAPI_FORCE_DEPS=1` exists. Requiring `node_modules/.bin/tsx` to exist before trusting the stamp is cheap insurance.

**The CI `paths:` gate misses three files that this workflow exists to protect.** `package.json` (every target is a wrapper over an npm script), `package-lock.json` (the dep-stamp step is specifically testing lockfile changes, and the last two commits on `main` are dependabot bumps of exactly that shape), and `.gitattributes` (the CRLF step asserts an invariant that file enforces).

**`--retry=1` is broader than the problem.** There is no `retry` in `vitest.config.ts`, so it applies to the whole suite, not the one network test. And the mitigation is one-sided: the existing Unit Test job runs the same suite with no retry, so if the stated cause is real it is that job that goes red. `traceContextPropagation.test.ts` is also not the only offender, `tracingIntegration.test.ts:16` hits a second live endpoint unmocked. Not yours to fix here, and I will raise stubbing both as its own issue, but the retry flag will hide the next real failure in this job too, so I would rather it came out.

## The web UI: dropping it, as you offered

Taking your offer. Revert the `package.json` change and take the container API-only. `src/webServer.ts` serves the SPA from `web-ui/dist`, which is both gitignored and dockerignored, so nothing in the container can produce it and `/` will 404 while `/api/*` keeps working.

That makes these dead:

| File | Remove |
|---|---|
| `package.json` | the `npm run web:build` addition to `"web"` |
| `docker-compose.yml:22-31` | the whole `web:` service, taking `PORT`, `ENSURE_WEB_UI` and the `WEB_PORT` mapping with it |
| `Makefile:81-83` | the `web` target |
| `Makefile:95-103` | the `COMPOSE_PROFILES=web` prefixes and their comment, inert with no profiled service left |
| `parksapi-entrypoint.sh:34-36` | the `ENSURE_WEB_UI` branch, nothing sets it any more |
| `dev_container.yml:60-70` | the web step. Leaving it in hard-fails CI: it asserts `/` returns 200, gets 404, and burns the full 5 minute curl loop first |
| `README.md:93`, `WEB_ADMIN.md:24,35` | the three `make web` UI pointers |

Keep the 8080 to 8888 corrections at `WEB_ADMIN.md:34` and `:174`. Those fix a genuine pre-existing error against `src/webServer.ts:21` and have nothing to do with the container.

For the docs, something like: the container serves the API only; the React admin UI is a host-side build, run `npm run web:build` on the host once and the container picks up `web-ui/dist` through the bind mount.

## Notes

- **The 80 count is right**, and thank you for catching it. 84 `@destinationController` occurrences under `src/parks/`, 4 of them prose inside doc comments (`te2.ts:154`, `fantawild.ts:25`, `parcsreunidos.ts:90`, `hfe.ts:102`), leaving 80 real decorators. While confirming it I found the destination table below `README.md:117` is worse than stale: 74 rows against 80 registered ids, missing 15, and listing 9 that are not registry entries at all (`cedarpoint`, `carowinds`, `knottsberryfarm`, `canadaswonderland` and the rest of that group are sub-parks behind the single `SixFlags` controller). Not yours, and please do not take it on in this PR, but line 117 claims the table equals `--list` output and it does not. I will raise it.
- **`--env-file /dev/null` is right, the stated reason no longer reproduces.** On current compose both quoted and unquoted multi-line values parse without error. The flag still earns its place, because without it a `.env` key named `TZ`, `WEB_PORT`, `UID` or `GID` silently wins the interpolation. Worth rewording the comment at `Makefile:12-13` so nobody deletes the flag after testing the claim and finding it untrue. The flag itself is applied on every path, including both podman recipes, and `RUN` is recursive so a `make COMPOSE=...` override propagates correctly.
- **The `--ignore-scripts` rationale does not carry to runtime.** `Dockerfile:18-19` argues `prepare: tsc` cannot run because there are no sources yet, which is true at build time but not in the entrypoint, where the bind mount has them. Net effect is that `dist/` is never produced by the container's install. Only `main: dist/index.js` consumers care, so it is minor, but the docstring for `make deps` could point at `make compile`.
- `docker-compose.yml:14` `:z` relabels the whole tree including `.git` on every start, and it persists beyond the container. Right call for Fedora, worth a word in the comment.
- `Makefile:98` says "Remove containers and networks" while the recipe passes `-v`. No named volumes today, so harmless.
- The write probe at `parksapi-entrypoint.sh:7-13` asserts a cause it has not tested: a read-only mount with a matching uid prints "bind-mounted from a host user this container is not mapped to" and advises the uid you already passed. Exit 1 is right, the message is not.
- `make cache-clear` with no ARGS clears the cache and then runs all 80 parks. Matches its docstring, but that is an expensive default for a target whose name suggests it only clears.
- Verified clean, for the record: the stamp is byte-identical between build time, runtime and the host; a partway `npm ci` failure aborts before the stamp is written, so it self-repairs; the `.gitattributes` lines cover all three new files; and the Dockerfile's layer and `USER`/`chown` ordering is correct.

Nothing here is a reason to start over. The container design is right, and once `.dockerignore` is default-deny and the project name is per-tree I am happy to take it.

BeyondVertical:
All three blockers and every should-fix are in, the web UI is out, and the branch is rebased onto current `main` (it predated #294/#296/#297). Six commits on top of the original thirteen, one per finding.

Verified on Docker 29.7.2 / Compose v5.4.0.

## Blockers

**1. `.dockerignore` is now default-deny.** `**/` on every pattern would only have closed the instances we thought to enumerate, so I took the better option: `*`, then back in `package.json`, `package-lock.json`, `tsconfig.json`, `vitest.config.ts`, `vitest.setup.ts`, `parksapi-entrypoint.sh`, `src` — and a re-exclusion of `**/*.har`, `**/testout_*`, `**/temp*` and `**/node_modules` inside `src`.

Reproduced your finding first with a nested checkout planted at `nested/inner/`: before, `cache.sqlite`, `.env` and an untracked `scratch-note.txt` all arrive in the context. After, the context is exactly those seven entries and nothing else; `find /app` in the built image comes back empty for all of them. The CI step plants the same files now instead of only checking the names we happened to list.

**2. `COMPOSE_PROJECT_NAME` per tree**, your line verbatim at `Makefile:23-26`. Tested the way you framed it — two trees both named `parksapi`, a container running in A, `make clean` in B:

```
before:  Container parksapi-parksapi-run-0aac15c0637b Removed     <- A's container
after:   projects are parksapi-7b322a40 / parksapi-f9916558, A survives
```

**3. `HOME` and the npm cache are in the image now**, and moving them was not enough on its own. `npm cache clean --force` empties `/tmp/.npm` but leaves the directory owned by the build uid, so `docker run --user 1001:1001` hit the identical EACCES one path along; node's compile cache (2.6 MB) has the same shape. Both are removed at the end of the install layer — `/tmp` is 1777, so npm recreates them for whoever runs. `docker run --user 1001:1001 -v <tree>:/app` now installs and stamps cleanly; forcing `HOME=/home/node` back reproduces your error exactly.

## Should fix

**Concurrent `npm ci`.** Serialised behind an `mkdir` lock, with the stamp re-checked after acquiring it so the waiter does not install a second time. Two containers on an empty tree: one `→ installing dependencies…`, one `→ another container is installing dependencies, waiting…`, one install, correct stamp, no lock left behind. Same test with the old entrypoint: both install at once. A lock left by a killed container is broken after 30 minutes (tested with a backdated mtime), a held one is waited out and picked up on release.

**`sha256sum` piped into `cut`.** Unpiped, `${sum%% *}`. Missing lockfile: before `want=[]` and rc=0, after rc=1 at the point of failure.

**Stale stamp.** `node_modules/.bin/tsx` must exist as well. Stamp correct + tsx removed → reinstalls; stamp correct + tsx present → skips; stamp wrong → reinstalls.

**CI `paths:` gate.** `package.json`, `package-lock.json` and `.gitattributes` added to both lists.

**`--retry=1` is out.** You are right that it was one-sided and far broader than the problem. The job runs the plain `make test`; `make test` keeps its `ARGS` passthrough because the other targets have it.

## The web UI

Dropped, per your table: the `package.json` change reverted, the `web:` service, `PORT`/`ENSURE_WEB_UI`/`WEB_PORT`, the make target, the `COMPOSE_PROFILES` prefixes, the entrypoint branch, the CI step and the three `make web` pointers. `package.json` no longer appears in this PR's diff at all. The 8080 → 8888 corrections stay. README carries your wording: the container serves the API only, the UI is a host-side `npm run web:build` and `web-ui/dist` is visible through the bind mount.

With the last profiled service gone the compose anchors had one consumer left, so they collapsed into a plain service block.

## Notes

- **`--env-file /dev/null` comment reworded.** You are right that the multi-line parse error no longer reproduces on current compose; the comment now gives the reason that does — a `.env` key named `TZ`, `UID` or `GID` silently winning the interpolation. Thanks for catching that before someone tested the old claim and deleted the flag.
- **`--ignore-scripts` at runtime**: `make deps` now says it does not build `dist/` and points at `make compile`.
- **`:z`**: the comment says the relabel covers the whole tree including `.git` and outlives the container.
- **`clean`'s docstring** matches the `-v` it passes.
- **The write probe** no longer asserts a cause. It states what it observed and offers both remedies, and the podman line prints `$(id -u)` literally — to run on the host — rather than echoing the container uid you already passed. Read-only mount with a matching uid and mismatched uid both exit 1 with the same honest message.
- **`make cache-clear`** I left alone: expensive default, but it does what its docstring says and narrowing it felt like a separate decision.
- **README table / destination count**: not touched here, as you asked. Tracked in #305.

## Runs

Every target, on the real tree: `list` (80), `test` (77 files, 1776 tests, no retry), host suite identical, `compile` (dist host-owned), `park`, `category CATEGORY="Alton Towers"` (1 park, correctly quoted), `health ARGS=efteling`, `cache-clear ARGS=efteling`, `shell` (TTY, uid 1000), `rebuild --no-cache` (npm ci genuinely re-runs), `clean`, `clean-all`. `find . -uid 0` empty throughout.

Fresh clone, deliberately named `parksapi` this time, no `.env` and no `node_modules`:

```
make build && make park PARK=efteling   -> 4/4
.env            0 bytes, host-owned
node_modules    host-owned, stamp matches the lockfile
find -uid 0     empty
git status      clean
image           parksapi-4a1ec80d-parksapi
make clean-all  removed only its own image; the main tree's survived
```

Podman is still untested — none on this machine — and the recipes are still marked as such.

cubehouse (commented):
Everything from the last round checks out. Default-deny `.dockerignore` with the planted-file CI step, per-tree `COMPOSE_PROJECT_NAME`, `HOME` and the npm cache moved into the image with both caches removed at the end of the install layer, the `mkdir` lock with the re-check after acquisition, unpiped `sha256sum`, the `tsx` presence test on the stamp, the three files added to the `paths:` gate, `--retry=1` out, and the web UI cleanly removed. The rewritten write-probe message and the `--env-file /dev/null` comment both say what is actually true now.

One thing left, and it is the fix from round two quietly not applying on a whole class of host.

## `sha256sum` does not exist on macOS

`Makefile:23`. macOS ships `shasum`, not GNU coreutils, so there is no `sha256sum` on a stock install. The `$(shell)` then produces nothing and `cut -c1-8` of nothing is empty:

```
$ PATH=<no sha256sum> make show
/bin/sh: 1: sha256sum: not found
OLD=[parksapi-]
NEW=[parksapi-c43d6a8e]
```

So every checkout on a Mac renders `COMPOSE_PROJECT_NAME` as `parksapi-`, which is a valid project name, which means all of them share it again. That is blocker 2 from the last round back in full, plus a `command not found` line on every `make` invocation. And it lands on the likeliest host for this PR's whole premise: someone with Docker Desktop and no Node.

Portable, and verified both ways in the transcript above:

```make
export COMPOSE_PROJECT_NAME := parksapi-$(shell printf '%s' '$(CURDIR)' | { sha256sum 2>/dev/null || shasum -a 256; } | cut -c1-8)
```

If `sha256sum` is missing the shell consumes no stdin before failing, so `shasum` gets the full input; the `2>/dev/null` swallows the not-found message. Same digest from both on a GNU host, so no project name changes for anyone already using this.

Worth a matching thought for the CI job, which only ever runs on `ubuntu-latest` and so cannot catch this class. Not asking for a macOS runner here, but the entrypoint's own `sha256sum` is fine either way since that one runs inside Debian.

Push that one line and I will take it.

cubehouse (approved):
Approving. The `sha256sum` fallback is in and it does what it needs to.

I ran the line rather than read it, on a GNU host and with `sha256sum` removed from `PATH` to stand in for macOS:

```
GNU host             NAME=[parksapi-c43d6a8e]
shasum-only (macOS)  NAME=[parksapi-c43d6a8e]
```

Same digest both ways, so no existing project name moves, and the not-found line is gone. `dev-container` and `build (24.x)` are both green on `e484b575`.

That was the last thing outstanding. Everything from the earlier rounds holds up on re-read: default-deny `.dockerignore` with the CI step planting the files rather than checking the names we thought of, per-tree `COMPOSE_PROJECT_NAME`, `HOME` and the npm cache in the image with both caches removed at the end of the install layer, the `mkdir` lock with the stamp re-checked after acquisition, unpiped `sha256sum`, the `tsx` presence test, and the `paths:` gate covering the files the job exists to protect.

One note on timing rather than content: I have merged #309 first, which takes the gate suite off `httpbin.org` and `jsonplaceholder.typicode.com` and blocks non-loopback DNS in `vitest.setup.ts`. Your job runs `make test` in-container, so it was the most exposed to the flake that made `--retry=1` look necessary in the first place. It inherits an honest suite now, and dropping the retry was the right call.

Thanks for working through three rounds of this. The container design was right from the start and the rework has been careful each time.

BeyondVertical:
Fixed, your line verbatim.

Reproduced it first with a `PATH` that has everything `make` needs except `sha256sum`, which is the shape of a stock macOS:

```
before, GNU host    COMPOSE_PROJECT_NAME=[parksapi-af6b8fa0]
before, no sha256sum
                    /bin/sh: 1: sha256sum: not found
                    COMPOSE_PROJECT_NAME=[parksapi-]

after,  GNU host    COMPOSE_PROJECT_NAME=[parksapi-af6b8fa0]
after,  no sha256sum
                    COMPOSE_PROJECT_NAME=[parksapi-af6b8fa0]
```

Same digest from both, so nobody already on this loses their project, and the not-found line is gone. `make list` on the real tree still builds under `parksapi-af6b8fa0-parksapi`.

The comment above the line now says why the fallback is there, so the next person to tidy it up knows an empty digest puts every checkout back into one shared project rather than merely looking untidy.

Agreed on CI: `ubuntu-latest` cannot catch this class, and a macOS runner for one `$(shell)` is not worth the minutes. The entrypoint's own `sha256sum` runs inside Debian, so it stays as it is.

PR #303, opened 2026-08-14
fix(walibi): merge CMS locales so rides missing from one feed still publish

lgnap:
## What this fixes

The CMS serves one attractions feed per locale and those feeds are maintained
independently, so they drift. Walibi Belgium's `nl` and `en` feeds both omit
VAMPIRE (`waitingTimeName` 34) while `fr` lists it: the ride had no entity and
no live data, even though `waitingtimes.v1.json` reported it open throughout.

Two changes:

1. **Walibi Belgium moves `nl` → `fr`.** Wavre is francophone and `fr` is the
   better-maintained feed — it is the only locale that lists VAMPIRE. This is
   what actually recovers the ride. It also fixes `FLASH-BACK` → `FLASH BACK`,
   `N'SOSO` → `N'Soso`, and moves `dining_cine-chill` from ~1 km outside the
   park to inside it. All 23 restaurant slugs are byte-identical between the
   two locales, so no restaurant id moves because of the locale change.
2. **`getAttractions()` merges the other locales in**, keyed on
   `waitingTimeName` (the only id stable across locales — title and path are
   both locale-specific). With the locale switch in place the merge no longer
   does anything for VAMPIRE; its live effect today is Le Galion and Tam Tam
   Aventure on Walibi Rhône-Alpes (entities 40 → 42, no live-data change since
   neither id appears in that park's wait-times feed). It is there so the next
   CMS gap is a non-event.

   The configured locale is authoritative for names and coordinates; a fallback
   can only ever *add* a ride the primary never listed, never overwrite one.
   A fallback locale that fails is skipped; the **configured** locale is not
   optional and its failure propagates, so a transient 5xx skips the cycle
   instead of publishing an empty park and pinning `[]` in the 24 h cache.

   Cost: 3 → 5 upstream calls per park per day.

## Entity id sanitisation — this is a rename, flag it

Restaurants have no stable id, so their entity id is the last segment of the
CMS path, which the CMS builds from the display name without sanitising it:
"Pêche & Mignon" arrives as `p-che-&-mignon`.

**Nothing was ever rejected in production.** `Destination.getEntities()` does no
id validation; the `/^[\w.-]+$/` check lives only in the dev harness
(`src/testRunner.ts:151`). Walibi Belgium is fully published today, bad id
included. What the `&` broke is `npm run dev -- walibibelgium`, not the data.

So `pathSlug()` now replaces **only** the runs the charset rejects:

```ts
segment.replace(/[^\w.-]+/g, '-')
```

Anything already legal is untouched by construction — including the trailing
hyphens the CMS emits (`stardocks-caf-`, `wild-rock-caf-`, both live on Walibi
Rhône-Alpes). No hyphen collapsing or trimming, which would rename those two
and could silently merge distinct slugs (`cafe--rouge` / `cafe-rouge`) into one
id.

⚠️ **One live id changes, deliberately:**

```
dining_p-che-&-mignon  →  dining_p-che---mignon   (Pêche & Mignon, Walibi Belgium)
```

Verified against the live feed: that is the only id in the park that moves; the
other 22 restaurants and all 16 attractions are byte-identical. The existing
wiki record needs repointing rather than being orphaned — @cubehouse offered to
handle that end.

## Other changes

- `getRestaurants()` gets `cacheVersion: 2` alongside `getAttractions()`, so the
  locale switch is atomic instead of serving `fr` attractions next to
  `nl`-sourced restaurant names and the stale `cine-chill` coordinate for up to
  24 h after deploy.
- Wait times matching no attraction in any locale used to be dropped silently.
  They are now reported — but summarised: one `console.warn` per poll listing
  only ids not already seen (~68 at Bellewaerde would otherwise be 68 lines), so
  the steady state is silent and a genuinely new CMS gap is a short line of its
  own. The primary locale is no longer listed twice.

## Tests

14 tests across `entityIds.test.ts` and `localeMerge.test.ts`. Verified by
mutation — each of these breaks the suite:

| mutation | test that fails |
|---|---|
| primary-locale error swallowed | `the configured locale failing propagates, rather than publishing an empty park` |
| hyphens collapsed/trimmed in `pathSlug` | 4 id tests, incl. `leaves a trailing-hyphen slug untouched` |
| `WalibiBelgium.culture` back to `nl` | `is configured on the locale that lists VAMPIRE` |
| merge keyed on `title` | `matches locales on waitingTimeName, not on title or path` |
| merge keyed on `path` | same |
| `fetchAttractions` ignores its `culture` argument | `requests each locale under its own path` (drives the real `@http` method against a local server) |
| fallback overwrites primary | `keeps the configured locale name when a fallback locale spells it differently` |

Full suite: 77 files / 1728 tests pass. `npm run dev -- walibibelgium`: 4/4,
41 entities (16 attractions incl. VAMPIRE), 16 live rows.

## Not in scope

`460e803c-6f87-442c-86b6-3cd80280beaf` is published twice by Walibi Holland on
`main` ("Walibi Express Station 1" and "Station 2"), so Station 2 has never
existed on the wiki. Pre-existing, unrelated to this PR — @cubehouse is raising
it separately.

### DIFF
--- a/src/parks/walibi/__tests__/entityIds.test.ts
+++ b/src/parks/walibi/__tests__/entityIds.test.ts
@@ -45,11 +45,9 @@ describe('WalibiBase.buildEntityList — restaurant ids from CMS paths', () => {
 
     const dining = entities.filter(e => e.entityType === 'RESTAURANT');
     expect(dining.map(e => e.name)).toEqual(['Pêche & Mignon', 'Crêpes Corner']);
-    // Only the `&` is replaced, so the hyphens around it survive:
-    // `dining_p-che-&-mignon` becomes `dining_p-che---mignon`. Ugly, but it is
-    // the rename with the smallest blast radius, and it is the id the wiki
-    // record has to be repointed to.
-    expect(dining[0].id).toBe('dining_p-che---mignon');
+    // The `&` and the hyphens glued to it collapse into one: this is the id
+    // the live `dining_p-che-&-mignon` record has to be repointed to.
+    expect(dining[0].id).toBe('dining_p-che-mignon');
 
     for (const entity of entities) {
       expect(entity.id).toMatch(/^[\w.-]+$/);
@@ -83,6 +81,22 @@ describe('WalibiBase.buildEntityList — restaurant ids from CMS paths', () => {
     const ids = entities.filter(e => e.entityType === 'RESTAURANT').map(e => e.id);
     expect(ids).toEqual(['dining_cafe--rouge', 'dining_cafe-rouge']);
     expect(new Set(ids).size).toBe(2);
+  });
+
+  test('leaves every slug the charset already accepts byte-identical', async () => {
+    // The guarantee that makes this safe to ship: sanitising touches only runs
+    // containing a rejected character, so a legal slug matches nothing. These
+    // are real slugs from the live feeds, hyphen shapes and all.
+    const legal = [
+      'crepes-corner', 'stardocks-caf-', 'wild-rock-caf-', 'cafe--rouge',
+      'shiva-s-food-truck', 'n-soso', 'javass-coffee-corner', 'fun-o-clock',
+      'cine-chill', 'ice_saloon', 'pizza.solo', '-taco', 'taco-',
+    ];
+
+    const entities = await stubbedPark(legal.map(slug => restaurant(slug, slug))).getEntities();
+
+    expect(entities.filter(e => e.entityType === 'RESTAURANT').map(e => e.id))
+      .toEqual(legal.map(slug => `dining_${slug}`));
   });
 
   test('a slug made entirely of rejected characters still publishes, rather than leaving a stale row', async () => {
--- a/src/parks/walibi/__tests__/entityIds.test.ts
+++ b/src/parks/walibi/__tests__/entityIds.test.ts
@@ -1,0 +1,60 @@
+import {describe, test, expect, beforeEach, afterEach} from 'vitest';
+import {WalibiBelgium} from '../walibi.js';
+import {CacheLib} from '../../../cache.js';
+import type {HTTPObj} from '../../../http.js';
+
+/**
+ * Restaurants have no stable id in the API, so their entity id is the last
+ * segment of the CMS path. The CMS builds that segment from the display name
+ * and does not sanitise it: "Pêche & Mignon" becomes `p-che-&-mignon`, and the
+ * `&` fails the harness id charset (`/^[\w.-]+$/`). Because validation runs
+ * over the whole list, that single restaurant made getEntities() throw — every
+ * Walibi Belgium entity, rides included, was rejected with it.
+ */
+const restaurant = (title: string, slug: string) => ({
+  title,
+  path: `/content/dam/wbe/nl/restaurants/${slug}-files/${slug}`,
+});
+
+function stubbedPark(restaurants: any[]): WalibiBelgium {
+  const park = new WalibiBelgium();
+  park.fetchAttractions = (async () => ({json: async () => []} as any as HTTPObj)) as any;
+  park.getRestaurants = async () => restaurants;
+  park.getWaitTimes = async () => [];
+  return park;
+}
+
+describe('WalibiBase.buildEntityList — restaurant ids from CMS paths', () => {
+  beforeEach(() => CacheLib.clear());
+  afterEach(() => CacheLib.clear());
+
+  test('sanitises a CMS slug the id charset rejects, without dropping the entity', async () => {
+    const entities = await stubbedPark([
+      restaurant('Pêche & Mignon', 'p-che-&-mignon'),
+      restaurant('Crêpes Corner', 'crepes-corner'),
+    ]).getEntities();
+
+    const names = entities.filter(e => e.entityType === 'RESTAURANT').map(e => e.name);
+    expect(names).toEqual(['Pêche & Mignon', 'Crêpes Corner']);
+
+    for (const entity of entities) {
+      expect(entity.id).toMatch(/^[\w.-]+$/);
+    }
+  });
+
+  test('leaves already-clean slugs untouched, so existing ids do not shift', async () => {
+    const entities = await stubbedPark([restaurant('Crêpes Corner', 'crepes-corner')]).getEntities();
+
+    expect(entities.find(e => e.entityType === 'RESTAURANT')?.id).toBe('dining_crepes-corner');
+  });
+
+  test('a slug made entirely of rejected characters yields no entity rather than a bare prefix', async () => {
+    const entities = await stubbedPark([
+      restaurant('???', '&&&'),
+      restaurant('Crêpes Corner', 'crepes-corner'),
+    ]).getEntities();
+
+    const dining = entities.filter(e => e.entityType === 'RESTAURANT');
+    expect(dining.map(e => e.id)).toEqual(['dining_crepes-corner']);
+  });
+});
--- a/src/parks/walibi/__tests__/entityIds.test.ts
+++ b/src/parks/walibi/__tests__/entityIds.test.ts
@@ -6,10 +6,19 @@ import type {HTTPObj} from '../../../http.js';
 /**
  * Restaurants have no stable id in the API, so their entity id is the last
  * segment of the CMS path. The CMS builds that segment from the display name
- * and does not sanitise it: "Pêche & Mignon" becomes `p-che-&-mignon`, and the
- * `&` fails the harness id charset (`/^[\w.-]+$/`). Because validation runs
- * over the whole list, that single restaurant made getEntities() throw — every
- * Walibi Belgium entity, rides included, was rejected with it.
+ * and does not sanitise it: "Pêche & Mignon" becomes `p-che-&-mignon`.
+ *
+ * That id publishes fine today — `Destination.getEntities()` does no id
+ * validation, and `dining_p-che-&-mignon` is a live record. What it breaks is
+ * `npm run dev -- walibibelgium`: the charset check at `src/testRunner.ts:151`
+ * runs over the whole entity list, so one bad restaurant rejects every Walibi
+ * Belgium entity, rides included. Sanitising it is therefore a *rename* of a
+ * live record, not a rescue.
+ *
+ * The flip side is that ids the charset already accepts must not move —
+ * including the trailing hyphens the CMS emits (`stardocks-caf-`,
+ * `wild-rock-caf-`, both live on Walibi Rhône-Alpes). Renaming those would
+ * mint new wiki records and orphan their history.
  */
 const restaurant = (title: string, slug: string) => ({
   title,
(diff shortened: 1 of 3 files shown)


cubehouse (changes requested):
Thanks for this, and for finding it at all. VAMPIRE genuinely is missing from the wiki today, so the gap is real and the locale explanation is right. I reviewed this across three passes (entity ids, the merge logic, and the tests against the live feeds) and verified the claims rather than reading them.

Three things need to change before it lands. The first is the one I'd fix even if nothing else changed.

## Blocking

### 1. A single upstream blip now publishes an empty park for 24 hours

`walibi.ts:182-189`. `fetchAttractionList` wraps the fetch in a blanket `catch { return []; }`, and it is used for the **primary** locale as well as the fallbacks. `fetchAttractions` sets no `retries`, so one timeout or 5xx returns `[]` rather than failing.

I stubbed one transient error and ran the same script on both branches:

```
origin/main   getEntities -> THREW: upstream 503
              getLiveData -> THREW: upstream 503

this PR       getEntities -> 25 entities, 0 attractions
              getLiveData -> 0 rows
```

On `main` the failure propagates, the cycle is skipped, and nothing is published. Here the same blip returns a clean, empty, wrong answer that reports itself as healthy. And because `getAttractions` is `@cache({ttlSeconds: 86400})`, that `[]` is pinned in SQLite for a day and shared by every process on that cache.

The partial case is the same shape with a quieter symptom: if Bellewaerde's `nl` fetch fails and `fr` succeeds, all 36 attractions come back French, `Dansende Ballonnen` becoming `Ballons Dansants`, and a Dutch park republishes in French for a day with no signal.

That is strictly worse than the bug being fixed here: VAMPIRE missing is one ride, this is all fifteen.

Fix is small, catch only for the fallbacks and let the primary propagate:

```ts
try {
  ...
} catch {
  if (culture === this.culture) throw;
  return [];
}
```

Worth noting your own test `'a locale that fails to load is skipped'` exercises only a fallback failing, so the dangerous path is precisely the one not covered.

### 2. `pathSlug` renames two entities that publish fine today

`walibi.ts:92`. The comment says "clean slugs come through unchanged, so existing ids do not shift". That isn't right: `.replace(/-{2,}/g,'-')` and `.replace(/^-|-$/g,'')` operate on characters the charset already accepts.

```
DRIFT "stardocks-caf-"   old="stardocks-caf-"   new="stardocks-caf"
DRIFT "wild-rock-caf-"   old="wild-rock-caf-"   new="wild-rock-caf"
same  "crepes-corner"    unchanged
```

Both are live on Walibi Rhône-Alpes right now:

```
RESTAURANT "dining_wild-rock-caf-" | Wild Rock Café
RESTAURANT "dining_stardocks-caf-" | Stardocks Café
```

A trailing hyphen is legal under `/^[\w.-]+$/`, so those ids were never broken. Renaming them doesn't rename anything downstream: the wiki mints new records and orphans the originals, losing their history. On a park this PR isn't trying to fix.

Sanitise only what the charset rejects, and anything already legal is untouched by construction:

```ts
const slug = segment.replace(/[^\w.-]+/g, '-');
```

The same two steps also make previously distinct slugs collidable, `cafe--rouge` and `cafe-rouge` both landing on `cafe-rouge`, `-taco` and `taco-` both on `taco`. A collision silently merges two venues into one entity and nothing detects it. Not triggered by today's data, but it goes away with the narrower replace. Separately, a segment that is entirely non-Latin or entirely punctuation sanitises to empty, returns `null`, and the entity is dropped rather than emitted, which leaves a stale row on the wiki rather than clearing it.

### 3. The premise needs correcting, and Pêche & Mignon is its own decision

Nothing was ever rejected in production. The `/^[\w.-]+$/` check lives only at `src/testRunner.ts:151`, inside the dev harness, and `Destination.getEntities()` does no id validation at all. Walibi Belgium is fully published today, bad id included:

```
Walibi Belgium  children: 39 | RESTAURANT: 23 | ATTRACTION: 15
  RESTAURANT "dining_p-che-&-mignon" | Pêche & Mignon
```

So what was broken is `npm run dev -- walibibelgium`, not the data. Please reword the PR description and the header comment in `entityIds.test.ts`, which says the same thing.

That matters because it changes what the `&` fix is. `dining_p-che-&-mignon` is a live record, so changing it to `dining_p-che-mignon` is a rename, not a rescue. It's a rename worth doing, but it needs a deliberate repoint of the existing record on our side rather than shipping quietly. Flag it in the description and I'll handle that end.

## Tests

The suite is genuinely load-bearing, 7 of the 8 new tests fail against pre-PR source. Two gaps:

- **`entityIds.test.ts:45`, the only test claiming id stability, uses `crepes-corner`.** No trailing hyphen, so it passes on both branches and misses the exact shape that breaks. `stardocks-caf-` is the case it wants.
- **Four mutations survive the suite.** Changing the merge key from `waitingTimeName` to `title` or to `path` is not caught, nor is inverting precedence so a fallback overwrites the primary, nor is making `fetchAttractions` ignore its `culture` argument, nor is putting `WalibiBelgium.culture` back to `'nl'`. The last one means the switch this PR actually ships is untested; the locale-merge tests set `park.culture = 'nl'` by hand.

The precedence one is worth a look: `'keeps the configured locale name when a fallback locale spells it differently'` passes under inversion only because the `en` fixture re-lists `FLASH-BACK`, so the last write happens to restore the primary. Drop it from the `en` fixture and the test does what its name says.

## Notes

- **The merge is redundant for the bug it was written for.** Moving Walibi Belgium to `fr` recovers VAMPIRE on its own; with the switch in place the merge adds nothing to that park. Its only live effect is Le Galion and Tam Tam Aventure on Walibi Rhône-Alpes, entities 40 to 42, no live-data change since neither id appears in that park's wait-times feed. Still worth having, just not for the reason stated.
- **The `nl` to `fr` switch is otherwise a small improvement.** Adds VAMPIRE, `FLASH-BACK` to `FLASH BACK`, `N'SOSO` to `N'Soso`, and moves `dining_cine-chill` from roughly 1km outside the park to inside it. All 23 restaurant slugs are byte-identical between the two locales, so no restaurant id moves because of the locale change.
- **`getRestaurants` (`walibi.ts:192`) has no `cacheVersion` bump** while its feed changes locale, so for up to 24h after deploy the park serves `fr` attractions next to `nl`-sourced restaurant names and the stale coordinate. `getAttractions` correctly went to 2. Bumping this one makes the switch atomic.
- **The orphan warning is 101 lines on the first poll** across three parks, and essentially all of it is permanent feed junk: 68 unmapped ids at Bellewaerde, 24 `closed_indefinitely` rows at Walibi Holland, 9 playground and `custom`-status rows at Rhône-Alpes. A genuinely new CMS gap arrives as one more line inside that. Worth either a count-and-summarise, or restricting it to ids that carry a real wait. Also `console.log` for a warning is unusual here, the other parks use `console.warn` 42 times against 4.
- The message prints the primary locale twice, `in any of [fr, nl, fr, en]`, because `mergeCultures` already contains `this.culture`.
- The comment at `walibi.ts:181` says a locale that 404s is skipped. The three locales the CMS doesn't serve return **200 with `[]`**, not 404, so that path never fires as described.
- Cost goes from 3 to 5 upstream calls per park per day, which is fine.

## Unrelated, and not yours to fix

While verifying the merge key I found Walibi Holland already publishes a duplicate entity id on `main`. `460e803c-6f87-442c-86b6-3cd80280beaf` is both "Walibi Express Station 1" and "Station 2", and the consequence is visible: the wiki carries only Station 1, so Station 2 has silently never existed. That is also why the merge-key mutation above goes unnoticed, since nothing checks for duplicate ids. I'll raise it separately.

## Summary

The locale switch, the merge, and sanitising the `&` are all worth taking. What I need is the `try/catch` narrowed to fallbacks, the sanitiser reduced to only the rejected characters, and the description corrected on the rejection claim. Happy to look again quickly once those are in.

lgnap:
All three blocking items are in, plus the two test gaps and the notes.

**1. Empty park on a blip** — `fetchAttractionList` now only swallows fallback
failures; the configured locale rethrows, so a transient 5xx skips the cycle as
it did before the merge existed.

**2. `pathSlug`** — reduced to `segment.replace(/[^\w.-]+/g, '-')`. Trailing
hyphens are legal and stay put (`stardocks-caf-`, `wild-rock-caf-` verified
unchanged), the `cafe--rouge`/`cafe-rouge` collision goes away, and a segment
can no longer sanitise to empty and drop the entity.

One correction to your snippet: with that regex only the `&` is replaced, so
the surrounding hyphens survive — the rename target is `dining_p-che---mignon`,
not `dining_p-che-mignon`. Checked against the live feed: it is the only id in
the park that moves, the other 22 restaurants and all 16 attractions are
byte-identical. Flagged in the description for the repoint.

**3. Premise** — description and the `entityIds.test.ts` header now say what is
actually true: nothing was rejected in production, the `&` broke
`npm run dev -- walibibelgium` via `testRunner.ts:151`, and sanitising it is a
deliberate rename of a live record. The commit message was reworded too.

**Tests** — the stability test now uses `stardocks-caf-`; `en` no longer
re-lists FLASH-BACK so the precedence test tests precedence. All four surviving
mutations are killed, verified by applying them one at a time: merge key →
`title` and → `path`, inverted precedence, `culture` back to `nl`, and
`fetchAttractions` ignoring its `culture` argument (that one needed a test
driving the real `@http` method against a local server, since stubbing
`fetchAttractions` hides the URL).

**Notes** — `getRestaurants` bumped to `cacheVersion: 2`; orphan wait times are
one summarised `console.warn` per poll listing only unseen ids (Bellewaerde's 68
become one line, steady state silent); primary locale no longer printed twice;
the 404 comment now says 200-with-`[]`; `getAttractions` doc no longer claims
the merge is what recovers VAMPIRE.

Full suite 77 files / 1728 tests green, `npm run dev -- walibibelgium` 4/4 with
41 entities.

lgnap:
Correction to my previous comment — the rename target is `dining_p-che-mignon`,
the id you wrote, not the `dining_p-che---mignon` I posted. Don't repoint
anything yet if you already started.

I had taken your snippet literally. `replace(/[^\w.-]+/g, '-')` only rewrites
the `&` itself, so the hyphens either side of it survive and the result is
`p-che---mignon`. To land on `p-che-mignon` the run has to absorb the hyphens
glued to the rejected character:

```ts
segment.replace(/-*[^\w.-]+-*/g, '-')
```

This keeps your constraint intact, and makes it structural rather than
empirical: the pattern only matches a run containing a rejected character, so
any slug already satisfying `/^[\w.-]+$/` matches nothing and comes through
byte-identical. Trailing hyphens (`stardocks-caf-`, `wild-rock-caf-`) and
doubled ones (`cafe--rouge`, distinct from `cafe-rouge`) are all untouched — the
`cafe--rouge`/`cafe-rouge` collision you flagged cannot happen either.

Verified against the live Walibi Belgium feed: `dining_p-che-mignon` is the only
id in the park that moves, the other 22 restaurants and all 16 attractions are
byte-identical, and no existing slug collides with it.

There is a new test pinning the guarantee itself — `leaves every slug the
charset already accepts byte-identical`, run over the real slug shapes from the
feeds. Three separate mutations of the sanitiser are now caught: dropping the
adjacent-hyphen absorb, collapsing hyphen runs globally, and trimming
leading/trailing hyphens.

Description updated. Full suite 77 files / 1729 tests green,
`npm run dev -- walibibelgium` 4/4 with 41 entities.

cubehouse (approved):
Approving. All three blocking items and both test gaps are addressed, and the correction you posted yourself on the sanitiser is the right one.

`segment.replace(/-*[^\w.-]+-*/g, '-')` gives the guarantee structurally rather than empirically: the pattern can only match a run that contains a rejected character, so any slug already satisfying `/^[\w.-]+$/` matches nothing. That is the property I was asking for, and it is now pinned by `leaves every slug the charset already accepts byte-identical` rather than by a single well-chosen fixture.

I verified the result on #307, which carries these three commits unchanged, against the live feeds:

| park | entities before | after | ids removed | ids added |
|---|---|---|---|---|
| Walibi Belgium | 40 | 66 | 1 | 27 |
| Walibi Rhône-Alpes | 40 | 46 | 0 | 6 |
| Walibi Holland | 59 | 59 | 0 | 0 |
| Bellewaerde | 53 | 53 | 0 | 0 |

`dining_p-che-&-mignon` to `dining_p-che-mignon` is the only id in the estate that moves, exactly as you said. I will repoint that record rather than let a second one be created.

The merge is doing real work on Rhône-Alpes as described, and the warning line is accurate on the live feed:

```
[WalibiRhoneAlpes] "Le Galion" (LE_GALION) is missing from the 'fr' feed, taken from 'en'
[WalibiRhoneAlpes] "Tam Tam Aventure" (TAM_TAM_AVENTURE) is missing from the 'fr' feed, taken from 'en'
```

20/20 tests green, harness 4/4 on all four parks, no duplicate ids introduced on any of them.

Merging this, then #307.

PR #307, opened 2026-08-15
Fix/walibi publish untagged attractions

lgnap:
Stacked on #303 — it needs that PR's `pathSlug()`, so the diff here shows #303's
three commits too. **Only the last commit is new**; the rest merges away with #303.

`buildEntityList` keyed attractions on `waitingTimeName`, so a POI without one was
dropped. Only rides with a queue board carry that field, and Walibi Belgium hands out
16 across 41 POIs — every one of them typed `"poiType": "ATTRACTION"` by the feed. The
other 25 are the whole family and kids side of the park: GRAND CARROUSEL, OCTOPUS,
TREE HOUSE, WAVE SWINGER, TAPIS VOLANT, the lot.

They published until 42008c71 ("switch to stable API IDs instead of slugified names"),
which added `.filter(a => a.waitingTimeName)` as a null-guard for the new id — the same
guard restaurants got in that commit, where it drops nothing because every restaurant
has a path. Nothing in the commit message or the code suggests the loss was intended,
and the file header still described the old `attr_xxx` scheme. It went unnoticed because
the guard is a no-op on the parks that tag everything: Walibi Holland is 39/39 and
Bellewaerde 36/36.

So fall back to the CMS path, exactly as restaurants already do. A tagged ride keeps its
wait-time id — those are live wiki records and must not shift. A POI with neither is
still dropped rather than minting a bare prefix. `pathSlug()` already sanitises the
segment, which matters here: the nl and en feeds path 4X4 ADVENTURE at a segment with a
literal space.

The 25 recovered rides all carry coordinates. They get no live data, since the
wait-times feed has no entry for them.

### Verification

`npx vitest run src/parks/walibi` — 20 tests green (5 new, in
`__tests__/untaggedAttractions.test.ts`).

Dev harness, all four parks on the live feeds, 4/4 each:

| park | attractions before | after |
|---|---|---|
| Walibi Belgium | 16 | **41** |
| Walibi Rhone-Alpes | 29 | **33** (its four Aire de Jeux) |
| Walibi Holland | 39 | 39 |
| Bellewaerde | 36 | 36 |

Rhone-Alpes' `Missing location: ATTRACTION:1` warning is pre-existing — it is already
there at 29 attractions, before this change.

### DIFF
--- a/src/parks/walibi/__tests__/entityIds.test.ts
+++ b/src/parks/walibi/__tests__/entityIds.test.ts
@@ -45,11 +45,9 @@ describe('WalibiBase.buildEntityList — restaurant ids from CMS paths', () => {
 
     const dining = entities.filter(e => e.entityType === 'RESTAURANT');
     expect(dining.map(e => e.name)).toEqual(['Pêche & Mignon', 'Crêpes Corner']);
-    // Only the `&` is replaced, so the hyphens around it survive:
-    // `dining_p-che-&-mignon` becomes `dining_p-che---mignon`. Ugly, but it is
-    // the rename with the smallest blast radius, and it is the id the wiki
-    // record has to be repointed to.
-    expect(dining[0].id).toBe('dining_p-che---mignon');
+    // The `&` and the hyphens glued to it collapse into one: this is the id
+    // the live `dining_p-che-&-mignon` record has to be repointed to.
+    expect(dining[0].id).toBe('dining_p-che-mignon');
 
     for (const entity of entities) {
       expect(entity.id).toMatch(/^[\w.-]+$/);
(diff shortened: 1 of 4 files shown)


cubehouse (approved):
Approving. The diagnosis is right, the fallback is the same one restaurants already use, and the recovered rides are real.

Verified against the live feeds, on a tree with current `main` merged in (so this ran through #299's orphan reporting as well):

| park | entities before | after | ids removed | ids added |
|---|---|---|---|---|
| Walibi Belgium | 40 | 66 | 1 | 27 |
| Walibi Rhône-Alpes | 40 | 46 | 0 | 6 |
| Walibi Holland | 59 | 59 | 0 | 0 |
| Bellewaerde | 53 | 53 | 0 | 0 |

The 25 recovered Belgian rides are the ones you named, and every one carries coordinates. No wait-time id shifts, no duplicate ids introduced on any of the four parks, no orphans in either direction. 20/20 tests, harness 4/4 on all four.

`dining_p-che-&-mignon` to `dining_p-che-mignon` remains the single id that moves, and I am repointing that record rather than letting a second one be created.

Two notes for later, neither of them blocking and neither of them yours to fix here:

**`culture` is now load-bearing for entity ids.** It was names only before this. I diffed the untagged slugs across locales on Walibi Belgium: 24 of the 25 are byte-identical between `fr` and `nl`, but the odd one out is not, and the charset is case-sensitive:

```
fr-only:  4x4-adventure
nl-only:  4X4-ADVENTURE
```

So changing a park's locale from here on renames whatever untagged rides differ. Worth a line on `this.culture` in the subclasses saying so, since the comment there currently reads as though the locale only affects display names.

**Two slugs that differ only in rejected characters collapse onto one id.** `a&b` and `a b` both sanitise to `a-b`. Not present in any of the four feeds today, and unavoidable given the sanitiser has to be idempotent on legal slugs. What makes it worth writing down is that nothing would catch it: `testRunner.ts` has no duplicate-id assertion, which is exactly why Walibi Holland has been publishing `460e803c-6f87-442c-86b6-3cd80280beaf` twice without anyone noticing. I will raise that check separately.

PR #313, opened 2026-08-16
fix(seaworld): stop publishing unmeasured rides as CLOSED

cubehouse:
> Stacked on #312. Base is `fix/seaworld-per-park-live-isolation`; merge that first and this retargets to main.

## Problem

The availability feed carries **three** wait-time states and we collapsed two of them. `Minutes: -1` is a sentinel for "no value"; `Status` separates a real closure from an absent reading.

| `Status` | `Minutes` | Meaning | We published |
|---|---|---|---|
| empty | `-1` | **no current reading** | ❌ `CLOSED` |
| empty | `>= 0` | operating | ✅ `OPERATING` |
| `Closed Temporarily` / `For The Day` / `Due To Weather` | `-1` | genuinely closed | ✅ `CLOSED` |

59 of 122 rides family-wide were mislabelled while parks were open, including Pantheon and The Big Bad Wolf at Williamsburg. Sesame Place San Diego is the extreme: all 7 rides unmeasured all day, so it published **zero operating attractions every operating day**.

Worse than the mislabelling: it **swallowed real closures**. A ride already showing `CLOSED` produces no visible change at the moment it actually closes, so the event never reaches consumers.

## Ground truth

`PoiAvailabilityEntityMapper.checkWaitTimeNull` in the vendor app (`com.seaworld.mobile`, two builds two months apart) discards the entire wait-time object when minutes are negative **and** status is blank, then renders the ride normally with no wait badge. It never shows such a ride as closed.

## Operating hours are load-bearing

The no-reading state says nothing about the ride, so hours arbitrate it. This is not decoration: **overnight the feed drops every ride into this state and strips all closure markers**. A 02:16 local sample had all 17 SeaWorld Orlando rides there, so without the check a shut park publishes its whole ride list as `OPERATING`.

Failure handling: hours undeterminable falls back to `CLOSED` rather than guessing, and a failed hours lookup never costs us the availability data already fetched. Both are tested.

## null, not undefined or omitted

`waitTime` is now always a number or `null`. Undefined serialised to `"STANDBY": {}`, which says nothing. `null` is the schema's documented *"queue exists but no current value is available"*, and that's accurate: rides sat unmeasured for hours, then reported real waits.

## Generic status matching

A 41-round census saw three closure strings, and `Closed Due To Weather` — absent from the first sample — was the **most frequent** (432 vs 287 vs 123). A hardcoded list would have shipped broken. There's a test asserting an unseen string still closes the ride.

## Evidence

41 rounds, 5207 observations, 12 park UUIDs, zero HTTP errors. Regression tests use observed sequences verbatim, e.g. Sunny Day Carousel: `norecord` 18:30Z → `Closed Due To Weather` 20:41Z → `0 min` 21:52Z.

## Verification

- 8 new tests fail against the old logic, all 59 pass with it
- Full suite 92 files / 1995 tests green
- Harness 4/4 on all 7 destinations. Overnight check at 02:20 ET confirms the hours gating: SWO dropped from 39 OPERATING to 24, Sesame San Diego from 12 to 5. Daytime behaviour is covered by unit tests against captured daytime payloads rather than a live sample, since parks were shut at commit time.
- Blast radius measured, not estimated: nothing outside `src/parks/seaworld/` imports `SeaworldDestination`; shows are on a separate path and unaffected; the 16 rides carrying genuine closure markers stay `CLOSED`.

## Also

`testRunner.ts` counted `waitTime !== undefined`, so `null` registered as a present wait time and a park where nothing is measured reported 100% coverage. Now counts actual values.

### DIFF
--- a/src/parks/seaworld/__tests__/seaworld.test.ts
+++ b/src/parks/seaworld/__tests__/seaworld.test.ts
@@ -102,14 +102,25 @@ const MOCK_PARK_DETAIL_SWO = {
   ],
 };
 
+// Field shapes here mirror real captured payloads. The three wait-time states
+// the feed actually uses are all represented:
+//   Status non-empty          -> genuinely closed
+//   Minutes >= 0              -> operating with a reading
+//   Minutes < 0, Status empty -> no current reading (the common case, 45% of
+//                                observations across a 41-round census)
 const MOCK_AVAILABILITY_SWO = {
   WaitTimes: [
     // Normal wait time
     {Id: 'ride-001', Minutes: 30, Status: '', StatusDisplay: null, Title: 'Ice Breaker', LastUpDateTime: '2026-04-01T10:00:00Z'},
-    // Closed ride (negative minutes)
-    {Id: 'ride-002', Minutes: -1, Status: '', StatusDisplay: null, Title: 'Closed Ride', LastUpDateTime: '2026-04-01T10:00:00Z'},
+    // No current reading: -1 sentinel with a blank status. NOT a closure.
+    {Id: 'ride-002', Minutes: -1, Status: '', StatusDisplay: null, Title: 'Unmeasured Ride', LastUpDateTime: '2026-04-01T10:00:00Z'},
     // Zero wait time (walk-on)
     {Id: 'ride-003', Minutes: 0, Status: '', StatusDisplay: null, Title: 'Walk On Ride', LastUpDateTime: '2026-04-01T10:00:00Z'},
+    // Genuinely closed, with the closure reason in both status fields
+    {Id: 'ride-004', Minutes: -1, Status: 'Closed Temporarily', StatusDisplay: 'Closed Temporarily', Title: 'Closed Ride', LastUpDateTime: '2026-04-01T10:00:00Z'},
+    // Closure string that did not exist when this park was first implemented —
+    // guards against anyone reintroducing a hardcoded list of known strings
+    {Id: 'ride-005', Minutes: -1, Status: 'Closed Due To Weather', StatusDisplay: 'Closed Due To Weather', Title: 'Weathered Ride', LastUpDateTime: '2026-04-01T10:00:00Z'},
   ],
   ShowTimes: [
     // Show with actual times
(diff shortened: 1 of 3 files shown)


cubehouse:
## Review round: 4 parallel reviewers, 6 fixes applied

Pushed as `a46447ee`. Findings that changed the code:

**Coercion.** `Number(wt.Minutes)` maps `null`, `''`, `'  '` and `[]` to `0`, which is finite and `>= 0`, so an absent field published OPERATING with a fabricated walk-on. The base-class sanitiser can't catch it because `0` is a valid wait time. Now only an actual number is trusted.

**Crash path.** `(wt.Status || wt.StatusDisplay || '').trim()` throws on a non-string field, and this loop runs *outside* the per-park try, so one malformed row would have taken down every park in the destination and undone #312's isolation. Coerced with `String()` first.

**Whitespace.** Same expression short-circuited on truthiness before trimming, so a whitespace-only `Status` won the `||` then trimmed to empty, discarding a real closure in `StatusDisplay`. Each field is trimmed separately now.

**Silent fallback.** `isParkOpenNow` returned false both for "shut" and for "hours published but nothing covers today" — the second silently flipping a park's unmeasured rides to CLOSED at midday. Both non-obvious cases warn now. Tri-state collapsed to boolean since the caller treated `null` and `false` identically.

**Dead code removed.** The negative-span guard was unobservable: an inverted block can't match a half-open interval anyway. Confirmed by mutation — deleting it failed no test.

**Two tests were ceremony.** The mock returned one park's availability for all three resort ids, so every row was processed three times and the last park's no-hours fallback overwrote the verdict. Deleting closure detection entirely left them green. Mock is per-park now, as the real feed is.

**DST flake.** The hours helper derived its window from `Date.now()`, and failed for the 60 minutes of the fall-back when the ambiguous wall clock resolved to the earlier offset. Clock is injected now. Green under UTC, Asia/Tokyo, Pacific/Kiritimati.

### Mutation results

Ten deliberate mutations, each now killed by at least one test. Five previously survived:

| Mutant | Before | After |
|---|---|---|
| ignore `Status`, read only `StatusDisplay` | survived | 1 test fails |
| remove `.trim()` | survived | 1 |
| remove the >25h glitch guard | survived | 1 |
| flip interval boundaries | survived | 1 |
| consider only `open_hours[0]` | survived | 1 |
| hardcoded closure list (no weather) | 1 test | 4 |
| delete closure detection | 3 tests | 6 |

### Deliberately not changed

A positive reading still publishes OPERATING regardless of the clock. Gating it on hours would suppress live data during an unpublished event (early entry, private party), and suppressing a reading we actually have is worse than the relic it prevents. The frozen-value problem is real and confirmed live (two coasters reading `waitTime: 0` at 02:51 local with the park shut) but it's pre-existing and needs staleness detection, so it's tracked separately rather than bodged in here.

### Corrections to the PR description above

- Census is **5002 observations**, not 5207 — that was the line count including blanks.
- **75 rows rewrite** on first deploy, not 59. The 16 genuine closures keep their status but still churn, because the payload goes `{}` to `{waitTime: null}`. 59 is the status-flip count in a single sampled round.
- The `STANDBY: null` precedent is **DLP and Phantasialand**, not SHDR or Universal, which use `SINGLE_RIDER`/`PAID_STANDBY`. DLP does the same hours-derived status with a null standby baseline.

### Deploy note

SeaWorld runs across multiple collector agents. Roll the SeaWorld-capable ones together, or the 10-minute full-resync backstop makes old and new builds re-assert conflicting snapshots and those rows flap until skew clears.

Full suite 92 files / 2005 tests. Harness 4/4 on all destinations.

PR #315, opened 2026-08-16
fix(plopsa): apply the park offset to show schedule times

cubehouse:
Show timeslots arrive as bare `HH:MM`. The schedule path built the entry time by string concatenation:

```js
openingTime: `${day.date}T${slot.start_time}:00`
```

leaving it with **no UTC offset** — despite the comment directly above claiming `constructDateTime` was used. The live path (`plopsaBuildShowtimes`, `plopsa.ts:170`) has always done it correctly; only the schedule path didn't.

## Impact

Anything parsing an offset-less timestamp as an instant resolves it in **its own** timezone, not the park's. Plopsa runs on Europe/Brussels, so every stored show time is served **one hour off in winter, two in summer**.

It also blocked the showtime backfill in ThemeParks/tpwserver#85. Comparing the stored schedule entry against the live observation for the same performance:

```
stored:   2027-01-03T12:15:00        ← resolved as UTC on the server
observed: 2026-08-16T12:45:00+02:00
```

Two hours apart, so every Plopsa day was skipped rather than repaired. 27 shows across Belgium and Deutschland publish live showtimes and were otherwise recoverable.

## Tests

Five new cases, all failing before the fix:

- the offset is present on both start and end
- an absent end mirrors the start, still with the offset
- an explicit end keeps its own offset
- a **winter** date resolves to `+01:00`, so a hardcoded `+02:00` can't pass
- the schedule path and the live path agree on the same performance **as instants** — the comparison the backfill actually makes

Reverting the fix to string concatenation fails all five. Full suite: 2014 passed across 93 files, typecheck clean.

## Scope

This corrects entries written from now on. Rows already stored keep their offset-less times, and the scraper only pushes forward-looking dates, so historical Plopsa entries won't be repaired by a re-push — the backfill has to interpret them instead. That's handled separately in tpwserver, and this PR is what makes the interpretation sound rather than a guess: it establishes that those strings were always meant as park-local wall clock.

### DIFF
--- a/src/parks/plopsa/__tests__/plopsa.test.ts
+++ b/src/parks/plopsa/__tests__/plopsa.test.ts
@@ -283,6 +283,78 @@ describe('buildSchedules feed independence', () => {
 
     expect(new Date(entry.openingTime).getTime())
       .toBe(new Date(live[0].startTime as string).getTime());
+    // Pinned absolutely as well: comparing only instants passes on a host whose
+    // own offset happens to match the park's, so on a CEST machine this test
+    // would survive the fix being reverted.
+    expect(new Date(entry.openingTime).toISOString()).toBe('2026-07-11T12:00:00.000Z');
+  });
+
+  /**
+   * constructDateTime throws on anything it cannot parse — a single-digit hour,
+   * a stray space, an empty string. The old concatenation never threw: it
+   * published one malformed entry and moved on. Without a guard, one bad slot in
+   * one show's feed rejects the whole buildSchedules run and takes the park's
+   * successfully-fetched operating hours with it, which is exactly the feed
+   * independence the tests above exist to protect.
+   */
+  test('drops a slot with an unparseable time instead of losing the whole run', async () => {
+    const badSlot = [{
+      ...showItems[0],
+      schedule_info: {
+        temporarily_closed: false,
+        schedule: [{
+          date: '2026-07-11',
+          timeslots: [
+            {type: 'open', start_time: '9:30', end_time: null},
+            {type: 'open', start_time: '14:00', end_time: null},
+          ],
+        }],
+      },
+    }];
+    const park = new ScheduleProbe();
+    park.fetchCalendar = async () => jsonResponse({schedule: {
+      '2026-07': {
+        '2026-07-11': {slots: [{
+          type: 'open',
+          start_time: '2026-07-11T10:00:00+02:00',
+          end_time: '2026-07-11T18:00:00+02:00',
+        }]},
+      },
+    }}) as any;
+    park.fetchEntertainments = async () => jsonResponse({items: badSlot}) as any;
+
+    const schedules = await park.buildSchedulesForTest();
+
+    // the park's own hours survive
+    expect(schedules.find((s) => s.id === 'plopsaland')?.schedule).toHaveLength(1);
+    // and so does the show's good slot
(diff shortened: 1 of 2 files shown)


cubehouse:
## Review round — one blocker fixed, one follow-up, one cross-repo finding

Three reviewers went over this. Two came back with real findings; both fixed here.

### Fixed: a single bad slot destroyed the whole run

`constructDateTime` validates nothing and throws `RangeError` on `'9:30'` (single-digit hour), a stray trailing space, `''`, `'25:00'`; `TypeError` on `undefined`. The concatenation it replaced **never threw** — it published one malformed entry and carried on.

Nothing caught it: only `fetchEntertainmentsAllLanguages` is wrapped, the slot loop is bare, and `getSchedules` has no handler. So one bad time in one show's feed rejected the entire `buildSchedules` run and **threw away the park's successfully-fetched operating hours** along with every other show — exactly the feed independence the surrounding tests exist to protect.

The tempting defence — the live path calls the same function in production without incident — is weaker than it looks: it parses only *today's* day (`find(d => d.date === date)`), while this parses every day the feed carries for every show. Orders of magnitude more values, including far-future provisional ones.

Now guarded per slot: a bad time costs that slot and nothing else. Two new tests, both failing before.

### Fixed: the instant test was host-dependent

`agrees with the live showtime path` compared only `getTime()`, so on a host whose offset matches the park's it passed **against the broken code**. Reverting failed 4 of 5 under `TZ=Europe/Brussels`, not 5 as the commit body claimed. Pinned absolutely; 6 now fail there.

### Exhaustive live sweep: the defect is isolated to this one path

All 197 parks and all 1,899 SHOW entities checked against the live API — 10,927 children enumerated, 14,530 timestamps scanned:

| layer | scanned | offset-less |
|---|---|---|
| park schedules (197 parks) | 10,172 | **0** |
| SHOW children (1,899, exhaustive) | 2,934 | **1,282** |
| ATTRACTION/RESTAURANT (sampled) | 1,424 | **0** |

Every hit is Plopsa, every hit is a SHOW. **25 shows, 641 entries** — 9 shows / 199 entries Belgium, 16 / 442 Deutschland. No other park has this class of bug.

Worth recording for future audits: `GET /entity/{parkId}/schedule` does **not** embed child schedules, so any park-level sweep returns a false all-clear. Children must be walked explicitly.

### Not fixed here — follow-up

Midnight-crossing slots produce a `closingTime` *before* the `openingTime` (`23:30` → `00:15` gives a −23h15 window). Pre-existing and identical in the live path, so not a regression from this PR, but it sits on the lines this PR touched. Four other modules roll the date forward (`fantawild.ts:257`, `europapark.ts:1145`, `disneylandparis.ts:1323`, `oceanpark.ts:864`). Filing separately rather than widening this diff.

### Cross-repo: this PR needs care at deploy time

Deploying this shifts published Plopsa show times by 1–2h. In tpwserver's merge, a stored offset-less time and the new offset-carrying one can land on **opposite sides of "now"**, so neither pairs and both are kept — permanently, since today's date is never pruned. Reproduced:

```
now    = 11:00 UTC
stored = 12:00:00        , 15:00:00
pushed = 12:00:00+02:00  , 15:00:00+02:00
result = 4 entries, not 2
```

Only *today* is at risk; past dates are started on both sides, future dates pending on both. Being handled tpwserver-side before this merges — details there.

Full suite 2016 passed, typecheck clean.

PR #316, opened 2026-08-17
fix(seaworld): stop publishing frozen wait times as OPERATING

cubehouse:
Closes card 185. Two headline coasters showed a walk-on wait at 3am.

## The feed never clears wait times — in two different ways

**It drops a ride to 0 and keeps refreshing that 0 all night.** Kraken, captured across closing:

```
21:01 ET  min=5  stamp 20:58  3m old
21:06 ET  min=0  stamp 21:03  3m old   <- park closes
...held, restamped every 5 minutes, until dawn
```

A staleness check cannot see this. The stamp is genuinely current.

**Separately, it freezes older values in place.** Four San Diego rides were still serving a 5-minute wait at midnight, stamped 13:07 — **651 minutes old**. A value check cannot see that either; the number is plausible.

One rule catches neither. What separates them is whether the park is running at all.

## The schedule must not be the only vote

Early entry, a private hire, an extra ticketed hour: the calendar says shut, the rides run, and suppressing a live reading then is worse than any relic.

So **a fresh positive reading anywhere in the park vouches for the whole park** and overrides the calendar. Safe because of what the overnight data shows:

| After 23:00 ET (all parks shut) | count |
|---|---|
| fresh + zero | 516 |
| stale + positive | 84 |
| stale + zero | 48 |
| **fresh + positive** | **0** |

Queues do not appear at a closed park — 0 of 648.

Both halves are load-bearing. "Any reading" fails on the 516 fresh zeros; "any positive" fails on the 84 stale positives. Only the conjunction separates them.

## The stamp is US Eastern for every park

Not the park's own zone. Across 1890 readings the median age is **3 minutes for all six reporting parks** and not one is future-dated. Read as park-local, San Antonio sits an hour ahead and San Diego three — which is what led #313's review to call the field timezone-inconsistent. It is consistent, just not local.

A test pins this with a **Pacific** park, where the two interpretations differ by three hours and only the correct one publishes. Without it the choice was unpinned, since every other fixture uses Orlando, where Eastern and local are the same.

Freshness is 60 minutes: in-hours ages cluster at a 3-minute median and 21-minute p90, while the stale positives served at midnight were 238 minutes old. Anything from ~40 to ~200 behaves identically.

## Verification

**Live, at 02:09 Eastern with every park shut:**

```
before:  OPERATING by entityType: { ATTRACTION: 2, SHOW: 22 }   <- Mako, Infinity Falls @ waitTime 0
after:   OPERATING by entityType: { SHOW: 17 }
```

Wait-time count went 2 -> 0 across all five parks checked. The remaining SHOW rows are card 184, untouched.

**Replaying the captured night** through the new rule: 1506 readings flip to CLOSED, 4510 keep publishing.

**Evidence base:** 12,549 observations over 103 rounds spanning every park's close, plus the earlier 5002-observation census. Zero HTTP errors in both.

**Mutation-tested.** Ten deliberate mutations each killed by at least one test, including dropping the operating check, ignoring freshness, accepting zeros as evidence, dropping either vote, widening the window to 24h, and reading the stamp as park-local.

Also fixed: `localFromFakeUtc` throws on an unreadable stamp, and this runs outside the per-park try, so a malformed timestamp would have taken down every park in the destination.

Full suite 2024 tests; green under UTC, Asia/Tokyo, Pacific/Midway.

### DIFF
--- a/src/parks/seaworld/__tests__/seaworld.test.ts
+++ b/src/parks/seaworld/__tests__/seaworld.test.ts
@@ -102,6 +102,22 @@ const MOCK_PARK_DETAIL_SWO = {
   ],
 };
 
+/**
+ * Stamps are US Eastern wall time with no zone suffix — that is the real shape
+ * for readings and closures (the fractional `…Z` form only ever appears on
+ * no-reading rows). These sit a few minutes before CLOCK_PARK_OPEN, which is
+ * 14:00 Eastern, matching the 3-minute median age measured in the wild.
+ */
+const FRESH_STAMP = '2026-08-15T13:57:00';
+/** Same shape, but hours old — the frozen-value case. */
+const STALE_STAMP = '2026-08-15T09:30:00';
+/**
+ * Fresh against CLOCK_PARK_SHUT (23:00 Eastern) rather than the open clock.
+ * Models the two things that actually happen after close: the feed refreshing a
+ * 0 every few minutes, and a genuinely live reading during an unlisted event.
+ */
+const FRESH_STAMP_NIGHT = '2026-08-15T22:57:00';
+
 // Field shapes match the five combinations seen across a 41-round census
 // (5002 observations, 2026-08-15). Note measured rows carry StatusDisplay: ''
 // while no-reading rows carry null — the API distinguishes them, so the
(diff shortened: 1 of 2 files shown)


cubehouse:
## Review round: 3 reviewers, 8 fixes

Pushed as `e93e02eb`.

### The regression two reviewers found independently

`parkIsOpen` is **null** when the hours lookup *failed*, which is not the same as knowing the park is shut — and I collapsed the two. Evidence does not reliably cover the gap:

```
SeaWorld San Antonio, Saturday afternoon
14:31 CT  evidence=0  readings=8/11  (every ride rolled back to a stale all-zero snapshot)
14:36 CT  evidence=0  readings=8/11
14:41 CT  evidence=0  readings=8/11
```

Park demonstrably open, bracketed by evidence rounds either side. With hours *also* unavailable in that window, all 11 rides would have published CLOSED. That turned a schedule outage from harmless-for-measured-rides into a whole park going dark. **When we don't know the hours, the reading now stands.**

### Freshness window 60 -> 90

`constructDateTime` resolves an ambiguous wall clock to the first pass, so during the DST fall-back hour every reading computes **exactly 60 minutes too old** — the threshold. The evidence vote would vanish family-wide, and `isParkOpenNow` fails identically at the same moment. In 2026 that hour is Halloween night, mid Howl-O-Scream.

Verified safe by measurement:

```
cap= 60min : 0 post-close readings qualify as evidence
cap= 90min : 0
cap=120min : 0
cap=180min : 0
```

### Also closed

- **Date-only stamp** parsed to local midnight and read as minutes old for the first hour after midnight — one such row could hold a shut park open through exactly the window this PR exists to fix.
- **Trailing-Z guard was case-sensitive** while the parser strips `/Z$/i`.
- **Malformed `WaitTimes`** (non-array, null elements, missing) threw out of `buildLiveData` and lost live data for every park in the destination. The neighbouring comment reasons about precisely this hazard while my new code added a second unguarded site.
- **Silent degradation**: if upstream changes the stamp format, evidence goes to zero with nothing in the logs. Now warns.

### The test hole

Every fresh fixture was **exactly 3 minutes old**, so **inverting the age subtraction passed all 79 tests**, and the freshness constant was pinned only to roughly [3, 810]. Fixtures now span the band: 30 min inside, 4 h outside, 2 min into the future for skew.

One test verified nothing: `survives an unparseable timestamp` put a valid stamp first, and `.some()` short-circuits, so the malformed rows were never parsed — it passed with the timestamp logic entirely removed. Bad rows now come first.

Eleven mutations, each killed by at least one test: sign flip, both window bounds, skew tolerance, undatable-as-evidence, closure guard, both stamp guards, park-local timezone, malformed-array guard, hours-unknown fallback.

### Corrections to the PR description above

| Claim | Actual |
|---|---|
| median age 3 min for all six parks | 3 min at five; **SeaWorld Orlando is 12** |
| Kraken "held it until dawn" | held 2h51m to the end of capture (23:57 ET) |
| safe window ~40-200 min | **40-180** — youngest post-close stale positive is 182 |
| stale positives "238 min old" | 182-651, median 597; 238 was the friendlier cluster |
| replay flips 1506 / keeps 4510 | correct numbers, but evidence-only over the **whole capture**, not the night. With a schedule vote: **852-1115** |
| "spanning every park's close" | Eastern/Central parks only; capture ends 20:57 local at the Pacific parks |
| in-hours p90 21 min | census only; freeze-day in-hours p90 is 195 min |

**Also worth knowing:** Sesame Place San Diego produced **zero** readings in 1008 observations, so the evidence path is structurally unreachable there and that park depends entirely on the schedule.

Full suite 2032 tests; 87 seaworld; green under UTC, Pacific/Kiritimati, America/New_York. Harness 4/4, and at 02:31 ET no attraction reports OPERATING at any park.

PR #321, opened 2026-08-19
fix(universal): stop reporting shows as OPERATING outside park hours

cubehouse:
## Summary

Universal show entities can remain live-looking through the overnight park closure because both the status field and the next operating day show list arrive before the park opens. This PR clock-gates ordinary shows against the schedule-bearing park venue while failing open whenever those hours cannot be trusted.

The original draft could incorrectly close ticketed-event shows while the daytime park was shut. That risk is now resolved:

- HHN shows are identified by the live feed category marker.
- HHN shows use the official ticketed-event window, not daytime park hours.
- Ordinary shows continue to use daytime hours even while HHN is running, so next-day slots do not become OPERATING after midnight.
- Missing, empty, mismatched, or failed event-calendar data fails open for HHN shows only.
- Hollywood now defaults to its own official HHN calendar and host park.
- The Tridion parser reads every calendar block, supporting Hollywood where an unbounded early-access block precedes the real event window.

## Schedule hardening

- Venue lookups run concurrently across the four Orlando parks.
- Non-array and empty schedule responses fail open.
- A malformed current-day row cannot be masked by a valid future row.
- A malformed early-entry timestamp falls back to a valid general opening.
- Explicitly closed schedules remain a confident CLOSED.
- Unmapped and non-schedule venues remain ungated.

## Live verification

Verified against the live feeds on 2026-09-04:

- At 23:58 PDT during HHN, The Purge: Dangerous Waters was OPEN with a 00:30 performance while the daytime venue was closed. The event-specific gate keeps it OPERATING.
- After the 01:00 event close, all 11 Hollywood HHN entities were CLOSED and no ordinary show remained OPERATING.
- Hollywood publishes 42 ticketed-event nights from the official calendar, including 2026-09-03 19:00 through 2026-09-04 01:00.
- During closed hours, Orlando returned 95 live rows with zero OPERATING entries.

## Verification

- Full test suite: 104 files, 2232 tests passed.
- Focused Universal suites: 90 tests passed.
- TypeScript build passed.
- Commit hook: typecheck, 110 affected tests, and 7 structural-gate tests passed.
- Live Hollywood and Orlando checks passed.

## Test plan

- [x] Ordinary show closes outside daytime park hours
- [x] Ordinary show stays operating during early/general hours
- [x] HHN show stays operating during a ticketed event
- [x] Ordinary next-day show remains closed during HHN after midnight
- [x] HHN show closes after the event window
- [x] Event-calendar empty/error paths fail open only for event shows
- [x] Hollywood multi-block calendar payload parses correctly
- [x] Malformed current day and malformed early-entry cases fail safely
- [x] Universal Orlando multi-park isolation
- [x] Full tests, build, and live closed-hours verification


cubehouse:
Holding this in draft. Found a real, confirmed risk after opening this PR: Halloween Horror Nights runs at both Universal Orlando and Universal Studios Hollywood as a wholly separate ticketed admission (day ticket ≠ HHN ticket, either direction), and the legacy `/venues/{id}/hours` schedule endpoint this fix clock-gates against shows zero signal of that evening window — confirmed against real data for 2026-08-28, Orlando's actual first HHN night this year (already locked into the schedule, only 9 days out at time of writing).

Unresolved: whether HHN-branded shows appear in `show-list.json` under the same day-park `venue_id` this fix gates. If they do, this fix would wrongly force them CLOSED during the live event.

Verifying live once HHN opens (~2026-08-28) rather than merging on a guess. Tracked as a gate on the internal board; will update here once resolved.

deano-000:
Thought it might be helpful for you to know team member events are Monday 24th/Tuesday 25th and premium scream Thursday 27th if you need to trial it.

PR #520, opened 2026-08-26
fix(energylandia): stop publishing Energy Pass entrance pins as rides

cubehouse:
The park's CMS added 14 `Wejście Energy Pass - <ride>` attraction documents. These are map pins marking the entrances to the paid skip-the-line queues ("Energy Pass" is the park's paid one-shot fast-track product, sold separately from admission and valid on 14 named rides). They are not rides: empty descriptions, no queue counter id, a shared placeholder catalogue number, and a `filterTags: ['energyPass']` marker driving the app map's filter. Each one was publishing as a real ATTRACTION entity reporting OPERATING all day.

This recognises the pins by name and keeps them out of both the entity list and live data, through the shared document filter so the two cannot disagree.

Matched on the name rather than the `energyPass` filter tag deliberately: tagging the real rides the pass covers is a plausible future CMS edit, and a tag-based filter would then silently drop 14 real coasters from the published entity list, which downstream is a deletion. No real ride will ever carry the product's own name.

Live verification: 103 raw active attraction documents → 14 pins stripped → 89 rides emitted, zero `Energy Pass` strings anywhere in the output, harness 4/4 green.

### DIFF
--- a/src/parks/energylandia/__tests__/energylandia.test.ts
+++ b/src/parks/energylandia/__tests__/energylandia.test.ts
@@ -15,6 +15,7 @@ import {
   buildShowtimes,
   showStatusFromShowtimes,
   resolveShowVenueId,
+  isEnergyPassEntrance,
   WEEKDAYS,
 } from '../energylandia.js';
 import {CacheLib} from '../../../cache.js';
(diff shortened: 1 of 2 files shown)


cubehouse:
Investigated whether the Energy Pass paid queues have a wait-time signal worth publishing. The wait feed carries 11 `FAST-PASS … LICZNIK` counter rows, but sampling every 10 minutes from before opening through peak midday shows all of them pinned at exactly 0 the entire time while adjacent standby counters climbed to 30–50 minutes. They are unwired operator counters, not a paid-queue measurement, so this PR deliberately publishes no `PAID_STANDBY` queue — emitting them would report a permanent zero-minute paid wait the park never measured. Documented in the `buildLiveData` comment; revisit only if one of those rows is ever seen non-zero.

PR #523, opened 2026-08-28
fix(walibi): publish the per-day hours override, not the season default

lgnap:
Walibi Belgium was publishing 10:00-20:00 for 28-31 August 2026 while walibi.be showed 10:00-18:00. The feed carries both numbers.

openinghours.v1.json holds two answers for a day's hours. openingHour/closingHour is the season default; customOpeningHourToDisplay is the per-day override, and it is what the calendar widget at the top of every page renders. buildSchedules() only ever read the structured pair, so a day whose hours were pulled forward after the season was set came out two hours long.

On those four days the override, the day's events asset (.../opening-hour-events/open-10-18h) and that asset's own title in the site's calendar feed ("10:00 - 18:00") all say 18:00. Only closingHour says 20:00 — it is the stale one. The site is inconsistent with itself for the same reason: its desktop nav badge reads the structured pair and announces "Ouvert aujourd'hui 10:00 - 20:00" next to a calendar saying 18:00.

The override is parsed strictly as HH:MM - HH:MM and nothing else. It is free text an editor types, and the event titles beside it show what free text does: "Halloween 10:00 - 22:00", and an open-10-22h asset titled "10:00 - 22:30". Anything that is not exactly two times leaves the structured hours in place — a shape we half-understand is worse than the default we already publish. Event slugs and titles are not parsed at all.

Scope, measured against the live feeds: of Walibi Belgium's 54 open days 44 carry an override and 40 of those agree with the structured pair, so this moves four days. Bellewaerde, Walibi Holland and Walibi Rhone-Alpes send the field empty on every open day, so the change is a no-op for them today and covers them the day it stops being one.

Fixtures are the real feed rows, including the December Walibi Winter day that carries no override at all.

### DIFF
--- a/src/parks/walibi/__tests__/openingHours.test.ts
+++ b/src/parks/walibi/__tests__/openingHours.test.ts
@@ -1,4 +1,4 @@
-import {describe, test, expect, beforeEach, afterEach} from 'vitest';
+import {describe, test, expect, vi, beforeEach, afterEach} from 'vitest';
 import {WalibiBelgium, WalibiHolland} from '../walibi.js';
 import {CacheLib} from '../../../cache.js';
 import type {HTTPObj} from '../../../http.js';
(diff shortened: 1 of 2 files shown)


cubehouse (approved):
This is a good change and the reasoning holds up. I checked it against the live feeds and the logic rather than taking the description on trust.

**What I verified**

- The two-answers problem is real, and `customOpeningHourToDisplay` is the one the site's own calendar renders.
- The strict `HH:MM - HH:MM` parse is the right call. Falling back to the structured pair on anything else means a free-text field an editor is still typing into can never produce a wrong schedule — only a less specific one. That is the correct direction to fail.
- Fallback is all-or-nothing, so a half-parsed override can never mix an overridden opening with a season closing.
- Full suite green (2149) on the branch, and your five cases cover the shapes that matter, including the day with no override at all.

**One thing to fix before merge, and it is only the comment**

```ts
// A park may legitimately close after midnight, so only reject hours no
// clock can hold.
if (openHour > 23 || closeHour > 23) return null;
```

The comment says after-midnight is supported, but nothing rolls the date, so an override of `10:00 - 01:00` parses happily and produces:

```
opening 2026-08-28T10:00:00+02:00
closing 2026-08-28T01:00:00+02:00
```

which closes nine hours before it opens.

To be clear about blame: **this is not something you introduced.** The existing structured path does exactly the same with a `closingHour` of `01:00`, so the behaviour predates the PR and you neither fix nor worsen it. What is new is the comment asserting the case is handled, which will send the next reader looking for handling that is not there.

Simplest resolution is to drop the "may legitimately close after midnight" sentence and say the bound is a sanity check on the hour field. If you would rather handle the rollover properly, that is a bigger change than this PR should carry — say the word and I will raise it separately so it does not hold this up.

**Two notes, neither blocking**

- The separator class accepts a hyphen and an en dash. Worth knowing that a park CMS elsewhere in this repo turned out to use both a hyphen *and* an en dash in the same document, so if you ever see an override fail to parse for no visible reason, that is the first thing to check.
- Nothing parses the event slugs or titles, which is the right restraint given `open-10-22h` is titled "10:00 - 22:30".

Approving on the strength of the change. Reword the comment and it is good to merge; happy to merge it for you once that lands, or leave it with you.

PR #543, opened 2026-09-06
fix(parcasterix): survive a CMS rebuild that renumbered and translated every POI

cubehouse:
## What happened

Parc Asterix rebuilt its Drupal between offline package **v1.1.335 and v1.1.336** (built 2026-09-02). Every POI id changed — 68 entities, **zero overlap** with the previous set — and the same release translated the whole POI list from French to English.

Wait times and showtimes are keyed by the same id, so the feed stayed internally consistent and nothing looked broken from inside. What broke is identity continuity for anyone holding the old ids.

| | before | after |
|---|---|---|
| Romus et Rapidus | `69` | `31303` → "Romulus and Rapidus" |
| Le Vol d'Icare | `636` | `31311` → "Icarus's Flight" |
| Goudurix | `11` | `31357` → "Justforkix" |

The package carries no stable alternative key. `map_id` is a park-map pin label — not unique across types, and reset whenever the map is reprinted — so the upstream node id remains the only usable id.

## Carry the other locales, for matching

The package ships one database per culture and, since the rebuild, a `title_fr` column in each. POI entries now carry every translation:

```json
{"en": "SOS Hieroglyphix", "fr": "SOS Tournevis", "es": "SOS Tornabis", "nl": "SOS Kissebis"}
```

**This does not change published output.** A `LocalisedString` name is flattened to `.en` downstream, and `en` was already the configured primary language, so every published name is byte-identical to before. The locales exist for the matcher below, which runs against parksapi's own output and can compare an entity against the name it *used* to be published under. They are also in place for whenever multi-locale names are supported end to end — but that is not today, and nothing here depends on it.

## Fix the migration matcher

The matcher existed for exactly this event and found almost nothing, because names acted as a gate that coordinates never got past.

- Names and coordinates are scored as **independent evidence and combined**. A park can keep every POI in place while changing every name, and that is a match.
- Alternate locales are scored alongside the primary name.
- Accents are **folded rather than deleted**, so `Aerodynamix` and `Aérodynamix` compare equal instead of merely similar.
- Candidates are scored **globally and assigned best-first**, so an early weak match can no longer claim an entity a later stronger one needed.
- The ranking score is **uncapped**. Neighbouring POIs routinely sit inside the "identical coordinates" radius of each other; clamping to 100 made them tie and let the wrong one win.

## Verification

Ground truth built independently, by diffing the last pre-rebuild package against the current one.

| | before | after |
|---|---|---|
| matched | 20 / 69 | **65 / 69** |
| correct against ground truth | — | **67 / 68** |
| missed | — | **0** |

The one remainder is a retired show paired with its replacement on the same stage, surfaced at 77% for review rather than auto-confirmed.

Regression-checked against parks whose ids have **not** changed — every entity must map to itself: Energylandia 170/170, Hersheypark 72/72, Walibi Holland 57/57, Liseberg 40/40, Walibi Belgium 63/64 (the one gap is a pre-existing duplicate row, which the matcher correctly refuses to double-assign).

## Tests

34 new tests across the matcher and the name builder, including the neighbour-tie case, the fully-translated-name case, the cross-entity-type guard, the distance rejection ceiling, and order-independence. Full suite: 2356 passing.

### DIFF
--- a/src/parks/parcasterix/__tests__/parcasterix.test.ts
+++ b/src/parks/parcasterix/__tests__/parcasterix.test.ts
@@ -1,0 +1,60 @@
+import {describe, it, expect} from 'vitest';
+import {buildLocalisedName, type POIEntry} from '../parcasterix.js';
+
+const poi = (title: string, titles: Record<string, string>): POIEntry => ({
+  drupal_id: 31303,
+  title,
+  titles,
+  latitude: 49.13675,
+  longitude: 2.573816,
+  _type: 'attraction',
+});
+
+describe('buildLocalisedName', () => {
+  it('returns every culture the offline package carried', () => {
+    expect(
+      buildLocalisedName(
+        poi('Romulus and Rapidus', {
+          en: 'Romulus and Rapidus',
+          fr: 'Romus et Rapidus',
+          es: 'Romus y Rápidus',
+          nl: 'Romus en Rapidus',
+        }),
+      ),
+    ).toEqual({
+      en: 'Romulus and Rapidus',
+      fr: 'Romus et Rapidus',
+      es: 'Romus y Rápidus',
+      nl: 'Romus en Rapidus',
+    });
+  });
+
+  // The French title is what makes an existing wiki entity still recognisable
+  // after Parc Asterix translated its whole POI list to English.
+  it('keeps the French title alongside the English one', () => {
+    const name = buildLocalisedName(
+      poi('Justforkix', {en: 'Justforkix', fr: 'Goudurix'}),
+    );
+    expect(name).toEqual({en: 'Justforkix', fr: 'Goudurix'});
+  });
+
+  it('falls back to a plain string when only one culture is present', () => {
+    expect(buildLocalisedName(poi('Toutatis', {fr: 'Toutatis'}))).toBe('Toutatis');
+  });
+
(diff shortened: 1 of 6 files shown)


cubehouse:
Added the retirement gate for Parc Asterix (`7fdcf32c`).

A show that ends its run leaves `paxSchedules` entirely rather than reporting no performances, and the offline package drops its POI row in the same release. `buildLiveData()` has nothing to key off, the collector is upsert-only, and the row freezes — four retired shows were reading OPERATING 8 to 24 days after their last write.

Opts into the shared gate from `destination.ts`. The 7-day default fits: shows drop out of `paxSchedules` on any day they do not perform, so the window has to clear an ordinary weekly cadence.

Two cases worth calling out, both covered by tests:

- **Seasonal closure** takes every show out at once for months. Force-closing them is the right answer, not a false positive, and attractions are unaffected because `paxLatencies` keeps listing them with `isOpen` false.
- **A gutted feed** parses to an empty array — `getPolling()` falls back to `[]` on a malformed response — and takes out far more than half the tracked entities, so the degraded-feed guard withholds.

**Limit, stated plainly:** this prevents recurrence, it does not clear the four rows already stranded. The gate only tracks ids it has seen in a live build, and those shows are absent from the package and the polling feed alike, so parksapi holds no record of them to retire. That one is a wiki-side action.

PR #560, opened 2026-09-10
fix(parcasterix): let the calendar alone decide a closed day's shows

cubehouse:
Follow-up to #558, which raised the fence rather than removing it. Caught live tonight by the sampler.

## What happened, 2026-09-10 (a closed day)

| Paris | open rides | ratio vs 41 |
|---|---|---|
| ≤17:39 | 4 | 0.098 |
| 17:42 | 5 | 0.122 |
| 17:48 | 7 | 0.171 |
| 17:54 | 8 | 0.195 |
| **18:03** | **11** | **0.268 — quorum crossed** |
| 18:06 | 12 | 0.293 |

Real headline rides: SOS Hieroglyphix, Osiris Convolvulis, By Toutatis!, The Tower of Edifis, The Flight of Ibis. Meanwhile `paxConfiguration` said `parkOpen: false`, the package calendar said type D, and the bill was the same out-of-hours default it had been serving for twenty-three hours.

Almost certainly a private evening event. The veto flipped, `all-dark` became `unknown`, and the build was about to republish that default programme as **eighteen performances on a day the park is shut to the public** — the state #558 cleared this morning, re-entered from the numerator instead of the denominator.

## The joint

`parkIsBusy` conflates *"the park is busy"* with *"the bill is about today"*. Both times it has misfired it was reporting truthfully; it was simply answering a different question. Eleven rides running for a corporate booking is no evidence that the show programme was rewritten.

So the calendar decides this on its own:

```ts
if (todaysHours.length === 0) return 'all-dark';
```

A private event now publishes its rides as OPERATING from `paxLatencies`, which is true, while its shows stay CLOSED, which is also true.

## Why not just raise the threshold again

Eleven of forty-one is a real fraction by any threshold that still catches a dead feed — and catching a dead feed is the only thing the quorum existed for. Raising it trades this failure for that one. Feed health is a separate question that `corroborated` already answers, so `MIN_OPEN_FRACTION` goes.

## The trade being made

The veto protected one case: a calendar that wrongly omits an operating day would force-close every show while they perform. That risk is bounded — the package calendar is the park's own and publishes weeks ahead, and it is cached 12h — and it fails toward *missing* information rather than *fabricated* information. The failure being removed has now fired twice in two days.

## Tests

Three tests named for the after-midnight tail of an event night were fixtured as ordinary 10:00–18:00 days, so what kept them silent at 01:30 was the veto, not anything about event nights — they were passing for the wrong reason. They now assert the real rule: at 01:30 on a day carrying no hours, nothing is running and the shows are dark. A genuine event night, whose entry closes after midnight, is covered by the mid-session test and reads its bill straight through, unchanged.

2465 pass. Live run mid-event: **12 rides operating, all 13 shows closed**.

### DIFF
--- a/src/parks/parcasterix/__tests__/showAbsence.test.ts
+++ b/src/parks/parcasterix/__tests__/showAbsence.test.ts
@@ -293,7 +293,15 @@ describe('Parc Asterix dark-show guards', () => {
   // The bill rolls at local midnight while a Halloween night still has an hour
   // to run, and a show physically mid-performance would otherwise be absent
   // from a bill for a day that has not started.
-  it('closes nothing in the after-midnight tail of an event night', async () => {
+  /**
+   * This was named for an event night and fixtured as an ordinary 10:00-18:00
+   * day, so what kept it silent at 01:30 was the ride-feed veto rather than
+   * anything about event nights. With the veto gone it asserts the real rule:
+   * at 01:30 on a day carrying no hours, nothing is running and the shows are
+   * dark. A genuine event night — one whose entry closes after midnight — is
+   * covered by the mid-session test, which reads its bill straight through.
+   */
+  it('closes the shows after midnight when no session is running', async () => {
     vi.setSystemTime(new Date('2026-10-17T23:30:00Z')); // 01:30 Europe/Paris
     const live = await stubbedPark(
       ATTRACTIONS,
@@ -301,7 +309,8 @@ describe('Parc Asterix dark-show guards', () => {
       [...ATTRACTION_POI, show(31483), show(31513)],
     ).getLiveData();
 
-    expect(live.filter((l) => l.status === 'CLOSED' && !l.queue)).toEqual([]);
+    expect(live.find((l) => l.id === '31483')).toEqual({id: '31483', status: 'CLOSED'});
+    expect(live.find((l) => l.id === '31513')).toEqual({id: '31513', status: 'CLOSED'});
   });
 
   // The bill served overnight is the last open day's and is not rewritten for
@@ -324,7 +333,13 @@ describe('Parc Asterix dark-show guards', () => {
 
   // The small hours are different: an event night's bill is still that night's
   // and still correct, so its performances must keep publishing.
-  it('keeps publishing performances in the after-midnight tail', async () => {
+  /**
+   * Same mislabelling: an ordinary day's calendar, read at 01:30. A bill entry
+   * timed 00:45 is a leftover from the default programme, not a performance in
+   * progress, and publishing it asserts a show on a day with no session. The
+   * real after-midnight case is the mid-session one below.
+   */
+  it('does not publish a leftover overnight performance on a dark day', async () => {
     vi.setSystemTime(new Date('2026-10-17T23:30:00Z')); // 01:30 Europe/Paris
     const live = await stubbedPark(
       ATTRACTIONS,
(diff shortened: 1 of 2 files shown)


cubehouse:
## Updated after review: attested closure vs unreadable day

Three reviewers went over this. Two independently blocked on the same defect, and it is real, so `4c0957cc` fixes it.

### What they found

Removing the ride-feed veto left `all-dark` reachable by a **calendar parse failure**, with nothing to catch it. The calendar is not fetched as hours — it is reconstructed by regex from free-form legend text, and `buildCalendarEntries` dropped a day whenever its type yielded no range. So `todaysHours.length === 0` meant three different things:

1. the park's legend says the day is closed,
2. the legend has hours whose wording the regex missed,
3. no row for that date at all.

Only the first is evidence. And `all-dark` is not silence — it closes `showIds ∪ scheduled`, so on a mis-parsed **open** day it publishes CLOSED over a bill actively naming those shows as performing. One reviewer measured it end to end:

```
31483  main: OPERATING + showtime 15:00   →  branch: CLOSED
31513  main: OPERATING + showtime 16:00   →  branch: CLOSED
```

**My original risk section was wrong on that path.** I claimed it fails toward missing information rather than fabricated; on this path it makes a positive false assertion.

### The fix

The park's own legend already carries the distinction, so it is now read rather than inferred from a gap. Dumped from the live package:

```
A  "10:00 a.m. to 6:00 p.m."   -> range
D  "Theme Park closed"         -> no clock time  = attested closure
J  "Daytime 9:00 ... Evening 7:00 p.m. - 1:00 a.m."  -> two ranges
```

A legend with **no clock time** is an attested closure and still drives `all-dark`. A legend **with** clock times that we could not pair into a range is a parser gap: the day stays out of the calendar, the shows publish nothing, and it **warns**. The time-token test is built from the same tokens as the range patterns, so the two can never disagree about what a time looks like.

All eight forward day types currently parse, and the 68 days of the 9 Nov – 18 Dec shutdown attest correctly, so the guard is silent today.

### Test coverage

`parseCalendarLabels` / `buildCalendarEntries` had **zero tests** despite being the sole authority for closing every show in the park. Now nine, including all ten current legends verbatim and the wordings most likely to appear next — an en-dash range, a French connector, an open-ended phrase — each asserted *not* to read as a closure.

Integration fixtures now have to state *why* a day carries no hours, so the two cases cannot be conflated in a test either.

7 of the new assertions fail against the previous head of this branch. 2474 pass. Live run on a closed day: 51 CLOSED, 4 OPERATING, no warning.

### Correction to the PR body

The body says three rewritten tests were "fixtured as ordinary 10:00–18:00 days". That over-generalises: two were, and the third (`stays silent in the after-midnight tail of an event night`) did have a genuine `TICKETED_EVENT` fixture — its flaw was that the clock sat 30 minutes *past* the night's close, so it never exercised the tail either. The conclusion is unchanged; the description was loose.

### Known gaps, not addressed here

- **No `getLiveData()`-level event-night coverage.** All `TICKETED_EVENT` fixtures are unit-level. Pre-existing, and the tail has a separate dating bug on both `main` and this branch (a 23:30 performance publishes with tomorrow's date).
- **Multi-session days collapse to `min(open)`..`max(close)`**, so the gap between two sessions reads as `read-bill`. Type J runs 11 days from 10 October with exactly that shape.
- **`tsconfig` excludes `__tests__`**, so changing a function's arity silently passes `undefined` to old call sites rather than failing the build. That bit me while writing this change.

Each deserves its own issue rather than more scope here.

Issue #561, opened 2026-09-11
Ocean Park: show entity ids change whenever the park rebrands a show


cubehouse:
## Problem

Ocean Park show entities are keyed on a slug of the programme title (`groupShowsBySlug` in `src/parks/oceanpark/oceanpark.ts`). The park rebrands its shows seasonally, so the same physical show emits a different entity id each season and the previous id goes silent:

```
Gala Of Lights: Sanrio characters’ Whimsical Celebration   -> show_gala-of-lights-...
Gala Of Lights – Pandastic Birthday Edition                -> show_gala-of-lights-pandastic-birthday-edition
Gala of Lights                                             -> show_gala-of-lights
```

One show, three ids. An entity id is the stable handle consumers key on, so a rename reads downstream as "one show retired, another appeared", and any history attached to the old id is stranded.

## Why the obvious fix doesn't work

The daily-schedule payload carries no identifier at all — just `title`, `timeSlot`, `locations`, `image`, `tab`. Attractions get CMS node UUIDs; shows get nothing.

The map feed (`map.oceanpark.com.hk/assets/data/shows.json`) does expose a short stable-looking `api_key` (`galaoflights`, `starmatch`, `neonls`), but it isn't a usable join key:

- It covers **5 of the 21** items on today's programme by exact name. The 16 it misses include every Animal Fun Talk and every feeding demonstration.
- It's stale: it still lists Chinese New Year 2026 entries and last summer's Pandastic shows.
- Two entries share the name "Roving Band" (`rovingband`, `rovingbandwf`) where the programme distinguishes them by a location suffix, so even the names that do match aren't reliably 1:1.

## Proposal

Slug the *core* title rather than the whole string: split on a colon or en/em dash and key on the head, leaving parenthetical suffixes intact.

```
Gala Of Lights: Sanrio characters’ Whimsical Celebration  -> gala-of-lights
Gala Of Lights – Pandastic Birthday Edition               -> gala-of-lights
Gala of Lights                                            -> gala-of-lights
Animal Fun Talk (Hornbill / Macaw / Cockatoo)             -> unchanged
Encounter with Star Performer (Rainforest Why Zone)       -> unchanged
Sanrio characters Meet & Greet (Summit) / (Waterfront)    -> unchanged, stay distinct
```

[...]

PR #562, opened 2026-09-14
fix(oceanpark): keep one show identity across a seasonal rename

cubehouse:
Refs #561

## Problem

Ocean Park rebrands its shows every season. No upstream feed gives a show an
id of any kind, so show entities are keyed on a slug of the programme title.
A rename therefore mints a brand new entity id and the old one goes silent.
One show, three ids:

```
Gala Of Lights: Sanrio characters' Whimsical Celebration
Gala Of Lights – Pandastic Birthday Edition
Gala of Lights
```

Downstream that reads as one show retiring and another appearing, and any
history on the old id is stranded.

## Approach: curate, don't infer

`SHOW_ALIASES` lists every title the park has published for one physical
show. A title belongs to that show if, and only if, it is listed. There is
deliberately **no rule that infers a family from a title nobody has vouched
for** — identity is data, not a heuristic.

An earlier revision of this PR did infer: it matched the *head* of an
unlisted title against a curated family name and adopted the canonical id.
That was dropped. Across 46 days of captured programme payloads the
inference never fired — one title touches the table, the bare `Gala of
Lights`, whose slug already is the canonical id, and no captured title
contains a subtitle separator at all. So the tests could prove it did not
break today's output and could not prove the rename path behaves, because
that path had never executed against real data. It would have run for the
first time in production, and its failure mode was silent: fold a genuinely
new production into an existing entity, or rename one. The park renames
these shows about twice a year; a two-line table edit each season is cheaper
than that risk.

What replaces it is a signal. A title whose head names a curated family but
is not itself listed publishes under its own id, as it always did, and logs
that it looks like a new edition and will strand the old id until someone
adds it. The separator scan survives only to decide whether to log; it
cannot change an entity id.

## Also: show coordinates were silently dead

The map exposes a URL on only 5 of 17 show entries, and the one that matters
has a URL slug that does not resemble its own name, so the URL join matched
nothing and **every** show fell back to the destination default. This indexes
the map's own `name` and `api_key` fields as well, skipping any key that
names more than one distinct position (the feed lists two different shows
both called "Roving Band"). Suppressing an ambiguous key is a shows-only
rule: the other map categories keep their long-standing last-wins behaviour.

Coordinate coverage on a current programme goes from 0/N to 4-5/N, the
remainder being shows the map has no unambiguous pin for.

## Feed validation

Three crash paths found in review, each one bad row away from taking out an
entire sync, since `groupShowsBySlug` runs inside `buildEntityList` with no
catch around it:

- a row whose `title` is missing or not a string
- a non-array `items` container (`body?.items ?? []` passes an object, a
  string, or a bare `0` straight into a `for...of`)
- a non-string `url` or `name` on a map row, which zeroed the coordinate map
  for every entity in the park

Plus: a projected coordinate is range-checked, not merely finite-checked; a
range keeps its start when only its end is unparseable (`18:00:00-24:00:00`,
the ordinary spelling for a midnight close, used to delete the performance
and publish the show CLOSED while it ran); and showtimes are deduped on
start time, keeping the later end, then sorted chronologically.

## Verification

Checked against 46 days of live programme data, 22 distinct titles:

- Entity ids and display names are **unchanged for every show the park is
  currently running**, on all 46 days, with **zero warnings**.
- A test iterates the real `SHOW_ALIASES` and asserts every family's id is
  the slug of one of its own titles, so a second family added with a
  mismatched id fails rather than silently never matching.

`tsc --noEmit` clean; full suite passing; 181 tests in this park, up from 96
on `main`.

## Review history

Three rounds of review against this branch. The first two found defects in
the change and then in its fixes — crash paths where an unvalidated feed
field escaped a builder with no catch, and order-dependences where feed row
order changed which showtimes were published. The third round measured the
machinery rather than arguing about it and most of it was removed: a
bracket-handling sub-system that bought one title shape no feed produces, a
venue check that never fired in 46 days and caused three defects, and
finally the identity inference itself. The last two commits each delete more
source than they add.

## Not covered by this change

The historical edition ids are already frozen upstream and this cannot close
or repoint them, since this destination does not run the retirement gate.
That remains a manual step, so #561 should not be closed by merging this
alone.


cubehouse:
Five parallel reviews over this branch turned up two crash paths and a regression, all introduced by the first commit. Fixed in aea01805, with a test for each.

**Crash: a malformed `locations` took down the entire entity list.** The venue check read the feed's shape without validating it, and `groupShowsBySlug` is called from `buildEntityList` with no catch around it. `locations` arriving as an object, a string, or an array holding `null` all threw a `TypeError` and rejected the whole build. Same class of failure as the range `timeSlot` that once took down every attraction wait time.

**Crash: a localised map `name` defaulted every coordinate in the park.** `slugify()` calls `.normalize()`, so a `{en, zh}` name threw, the caller's catch zeroed the coordinate map, and every attraction, restaurant and show fell back to the destination default. One row of seventeen was enough. `name` was never read before this PR, so the blast radius was new.

**Regression: the venue check fragmented the family it was protecting.** A bare family title's slug already *is* the canonical id, so refusing it left the row colliding with its own family and one of the two was dropped, depending on feed order. Upstream refining a zone id would have produced exactly the churn this change exists to abolish. The venue now guards only the head-matching heuristic, never a title a human has vouched for, and refuses a multi-zone row only when none of its zones is the curated one.

Also fixed from review:

- A curated title now outranks an unvouched one for a merged show's display name. Previously a members' backstage tour that merely heads onto the family could rename the flagship show, decided by punctuation codepoint.
- Showtimes dedup on the start alone, keeping whichever entry carries an end time. Merging joins the programme tab's bare start times to the event tab's ranges, which are two spellings of one performance; the old key published 19:00 twice and then dropped one the moment the show began.
- Showtimes are sorted chronologically, matching the other two parks that merge showtimes.
- The subtitle separator covers the fullwidth colon and the remaining Unicode dashes, and normalises before scanning. Those silently minted new ids.
- A separator inside brackets no longer splits a title.
- An ambiguous show is suppressed in the URL namespace too. Withholding only the name key left the bare URL slug free to serve the wrong Roving Band, so the guard did not achieve what its own comment described.
- Key ambiguity is judged by distinct position and after the coordinate filter, so a verbatim duplicate row no longer costs a show its pin.
- Non-finite pixel coordinates are rejected in the URL path as well; it admitted `NaN` and published a null latitude.
- The map key travels on the group rather than being re-resolved against the module table.
- A subtitled title belonging to no curated family now warns. A stale table otherwise degrades in silence, which is the failure this whole change exists to stop.

Verification: `tsc --noEmit` clean, full suite 2531 passing (153 in this park, up from 120). Mutation-tested at 20 deliberate breakages, 18 killed; the two survivors are provably unreachable given the curated early-return and are commented as such. Entity ids re-checked against the same 46 days of live programme data and remain byte-identical to `main` for every show the park is currently running.

Correction to the original description: the map exposes a URL on 5 of 17 show entries today, not 4. It does not change the conclusion, since none of those URL slugs matched a show before this change.

Known and out of scope here: the historical edition ids are already frozen upstream and this cannot close or repoint them, as this destination does not run the retirement gate. That stays a manual step, so #561 should not be closed by merging this alone.

cubehouse:
Third review round. This one measured the machinery rather than arguing about it, and the result is a commit that **removes more source than it adds** (-182/+146 in `oceanpark.ts`).

**The venue check is gone.** It never fired once across 46 days of captured programme data, and it cost three defects in two rounds: it fragmented the curated families it existed to protect; pooling zones to compensate turned absent metadata into a refusal, losing the canonical id; and pooling also made a row's id depend on which other rows shared its payload — which matters because `buildEntityList` and `buildLiveData` fetch the schedule independently and can read different cache generations. The zone a row names is upstream prose; the title is the identity. A surprising zone is now a log line, which is what a curated title already got.

**The bracket sub-system is gone.** Measured against the corpus and 2592 generated titles, four interlocking pieces bought correct handling of exactly one shape — `Family ( - Edition)` — that no feed has produced. Three of the four existed only to patch the first: the bracket rule could hide a separator behind a typo'd parenthesis, so a rescan undid it, so a suppression quietened the rescan, so a balance check un-suppressed the suppression. `titleHead` is now a regex and two lines, and a bracketed title still never folds into a family.

Also fixed from this round:

- The `items` container was unvalidated while its rows were not: `body?.items ?? []` passed an object, a string or a bare `0` into a `for...of`, and the throw landed after `buildLiveData`'s per-source catch had run.
- `parseShowTimeSlot` discarded a good start when only the end was unparseable. `18:00:00-24:00:00`, the ordinary spelling for a midnight close, deleted the performance and published the show CLOSED while it ran.
- A range now matches a *run* of dashes, as titles already did, so a doubled hyphen typed for an em dash does not cost the whole slot.
- A family's own bare name now ranks as vouched when choosing a display name, so an unvouched edition cannot take it on a lexical tie.
- **Non-show map categories get last-wins back.** Suppressing an ambiguous key is a shows-only rule; `animals` and `shops` already ship duplicate URL slugs, so an attraction would have lost its pin the day the feed duplicated one. That was a regression this PR introduced outside its own scope.
- A show can again reach a pin filed under another category as a last resort.
- The curated map key outranks a live name match again, and a disagreement between them is logged rather than resolved silently in either direction. Round two and round three each argued for the opposite order; neither is safe, so the clash is now surfaced.
- A projected coordinate is range-checked. `Number.isFinite` accepts a sentinel pixel that lands at latitude -934 and nothing downstream validates it.

Comments that asserted unobserved feed behaviour as fact now say what was actually measured, and the ones written for a reviewer rather than a maintainer are gone. Two tests named "every …" that hardcoded a single literal now iterate the real table, and five near-duplicates were dropped.

Verification: `tsc --noEmit` clean, full suite 2575 passing, 197 in this park. Entity ids and display names identical to `main` for every show the park currently runs, across all 46 days, with zero warnings.

Correction to earlier notes on this PR: "byte-identical to main" was too strong — ids and display names are identical; the internal group objects carry new fields. And the round-two summary understated its own size.

PR #568, opened 2026-09-18
fix(universal): an event-tagged show can also perform in the daytime

cubehouse:
## The gap

`category: hhn` marks a show as belonging to Halloween Horror Nights. It does not mean the show only ever runs during it — but the gate treated those as the same claim. For an hhn-tagged show the day-park verdict was replaced outright by the event verdict:

```ts
if (show.category?.toLowerCase() === 'hhn') {
  parkOperating = /* the event verdict */;   // the day verdict is discarded
}
```

Hollywood's "Meet HamiKuma" is tagged `hhn` and, sampled from the live feed, carries ENABLED performances from 13:30 PT straight through to 20:25 PT:

```
13:30  15:00  15:30  16:00  16:30 | 17:00 admission | 17:30 … 20:25
```

The day park runs 08:00-18:00 and the event admits from 17:00, so five of those slots happen in an open day park, before the event starts. All five were gated on an event that had not begun.

## Why it was not obvious

The 30-minute performance-underway rule tiles a run of slots spaced exactly 30 minutes apart, so most of the afternoon looked fine. That spacing is an accident of this season's scheduling, not a guarantee — the 13:30 slot has no successor for 90 minutes, and the gap published CLOSED beside a show with five performances still to come.

## The fix

Take the union: a guest can reach the show if EITHER the day park is open or the event is admitting.

The day arm requires a future performance inside today's day-park window. That is the exact mirror of the rule already applied to ordinary shows performing inside the event, and it is what keeps the original design intact — a maze carries no daytime slot at all, so an open day park says nothing about it and it stays gated on the event alone.

## Verification

Values in the fixtures were sampled from the live feed today, not invented: the slot list and `category` from the show feed, the 17:00 admission from the event calendar, the 08:00-18:00 window from the venue schedule.

- 8 new tests in `showClockGate.test.ts`. Two failed before the change (the 14:30 PT gap), six passed before and after — they are the guards: the maze stays CLOSED, an evening-only show is not unlocked at noon, the early-access arm still admits, the event arm still admits, and the overnight case still closes.
- Full suite: 2604 tests across 113 files, all passing.
- `npm run dev -- universalstudios`: 4/4.

One thing is not yet proven in production: the live confirmation has to be taken between 14:00 and 17:00 PT, which is the only window where the day park is open, the event is not, and the slots exist. That sample is scheduled for tonight and the result will be added here.


cubehouse:
## In-window confirmation, sampled live

Five samples across 13:00-14:20 PT on 2026-09-18, against the live feed, day park open 08:00-18:00 and the event admitting from 17:00:

| PT | Meet HamiKuma (`hhn`, daytime slots) | The Purge (`hhn`, evening only) | Death Eaters (`hhn`, no slots) |
|---|---|---|---|
| 13:00 | OPERATING | CLOSED | CLOSED |
| 13:20 | OPERATING | CLOSED | CLOSED |
| 13:40 | OPERATING | CLOSED | CLOSED |
| 14:00 | OPERATING | CLOSED | CLOSED |
| 14:20 | OPERATING | CLOSED | CLOSED |

HamiKuma's slots today are `15:00 15:30 16:00 16:30 17:00 17:30 …`, four of them before admission.

The part that matters: **every one of those samples is before the first slot of the day at 15:00**, so no performance has begun and the 30-minute underway rule cannot be what is holding the row open. The day arm is the only thing producing OPERATING. Before this change the same instants resolve to CLOSED, because the event verdict had replaced the day-park verdict and the event does not admit until 17:00.

The two guards hold at the same instants. The Purge performs only in the evening and Death Eaters carries no `show_times` at all, so neither has a future slot inside the day window, and an open day park does not reach either.

A sample at 04:58 PT, park shut and the event long over, reads CLOSED, so the overnight case is intact.

Sampler continues every 20 minutes to 17:59 PT, which will cover the tiled 15:00-17:00 band and the handover to the event arm at 17:00.

cubehouse:
Live confirmation, sampled on production every five minutes from 12:06 to 17:12 local on 2026-09-18 — the window this PR said it needed.

**Meet HamiKuma**, Universal Studios Hollywood. Park schedule that day:

```
EXTRA_HOURS    08:00 - 09:00
OPERATING      09:00 - 18:00     <- day park
INFO           17:00 - 19:00     (Halloween Horror Nights Early Access)
TICKETED_EVENT 19:00 - 02:00     (Halloween Horror Nights)
```

Performances: `15:00, 15:30, 16:00, 16:30, 17:00, 17:30, 18:30, 18:55, 19:25, 19:55, 20:25, 20:55`.

The first four sit entirely inside the day park and hours before the event opens.

## What the API published

```
time    status      perf within 15m   performances still to come
12:06   CLOSED      no                12
  ... unchanged for nearly three hours ...
14:47   CLOSED      YES               12
14:52   CLOSED      YES               12
14:57   CLOSED      YES               12
15:02   OPERATING   YES               11
  ... OPERATING for the rest of the sample ...
17:12   OPERATING   no                 7
```

Totals: 62 samples, 15 with a performance underway or within fifteen minutes, **3 of those published CLOSED**.

## What it shows

The show read `CLOSED` from 12:06 through 14:57 with twelve performances still to come, and flipped only at 15:02 — after the 15:00 slot had already begun. The event it was gated on opened at 19:00, four hours later.

It also locates the masking described in the PR body. The 30-minute performance-underway rule rescues the status *once a slot has started*, which is why the defect shows on the approach to 15:00 and not across the whole afternoon: three samples rather than fifteen. That is why this looked like an occasional glitch rather than a systematic gating error, and why the fixture evidence was the clearer signal.

## Note

Sampled through `GET /v1/entity/{parkId}/live` on the deployed build — so this is the behaviour **before** this PR, captured while it sat unmerged. Nothing here tests the fix; it confirms the defect the fix addresses.

cubehouse:
## Full window complete, 2026-09-18

Sixteen samples, every 20 minutes from 13:00 to 17:40 PT, against the live feed. Day park 08:00-18:00, event admitting from 17:00.

**Meet HamiKuma** (`hhn`, slots from 15:00) reads OPERATING at all sixteen. The three bands matter separately:

| Band | What carries it |
|---|---|
| 13:00-14:40 | the day arm alone. No slot has begun, so the underway rule cannot apply |
| 15:00-16:40 | day arm, with the underway rule incidentally tiling alongside it |
| 17:00-17:40 | handover to the event arm at admission |

Before this change the first band resolves to CLOSED, which is the bug.

**The Purge: Dangerous Waters** is the control, and it is the cleanest evidence here. Also `category: hhn`, also at the host venue, but its slots are 20:00 / 22:00 / 23:15:

```
04:58 - 16:40 PT   CLOSED       day park open for most of it, and it stays shut
17:00 PT           OPERATING    admission
17:20, 17:40 PT    OPERATING
```

Thirteen consecutive samples with the day park open and the row correctly CLOSED, flipping at exactly the admission boundary. That is the day arm declining to fire for a show with no performance inside the day window, measured rather than asserted.

**Death Eaters Encounter** carries no `show_times` at all and reads CLOSED at all sixteen.

So both halves hold on live data: an hhn show that performs in the daytime is no longer gated on an event that has not started, and one that does not is still gated exactly as before.

The overnight case is unchanged too, from the 04:58 PT sample with the park shut and the event over: CLOSED.

PR #576, opened 2026-09-19
test: make the park edge-case tests exercise the parks

cubehouse:
`src/__tests__/parkEdgeCases.test.ts` had 8 blocks and 43 tests. **One block tested real code.**

The rest asserted against values written in the test itself: a status table looped against a regex of the four valid statuses, a reimplemented HTML cleaner, reimplemented schedule regexes, and an object literal checked for the property it had just been given. None could fail whatever the parks did.

```ts
const eftelingStates: Record<string, string> = { 'storing': 'DOWN', … };

for (const [state, expected] of Object.entries(eftelingStates)) {
  test(`state "${state}" maps to ${expected}`, () => {
    expect(expected).toMatch(/^(OPERATING|DOWN|CLOSED|REFURBISHMENT)$/);
  });
}
```

The test name is a claim about Efteling's code. The assertion is that the string `'DOWN'` matches a regex containing `DOWN`.

## Two of the tables were untrue

**HFE.** The table claimed `TEMPORARILY CLOSED` and `TEMPORARILY DELAYED` map to DOWN. Both are gated on `parkIsOpen` and read CLOSED once the park shuts, which is right: DOWN claims something is wrong with the attraction, and outside operating hours nothing is wrong. A reader of those tests would conclude this park reports DOWN overnight.

**Phantasialand.** The table described a string status mapper. That park has no such thing. Status comes from a boolean (`phantasialand.ts:545`, `:553`) and grepping the module for `DOWN` returns nothing. The block documented a capability that does not exist. Removed rather than replaced; the gap it implies is filed separately.

## And one inline copy was right where the shipped code was wrong

The schedule block rewrote the Parques Reunidos regexes and, on the Dutch format, applied an afternoon adjustment the real parser does not have:

```
test:     '10 tot 5u' -> close 17:00
shipped:  '10 tot 5u' -> close 05:00
```

So `parseTimeRange` returned a park closing five hours before it opened and published that inverted window as OPERATING, while a green test said otherwise. Dutch drops the meridiem, so a bare closing hour at or before the opening rolls to PM. Fixed.

**Not observed live.** All 180 of Bobbejaanland's currently published schedule entries are well formed, so upstream is serving a shape that misses this branch. Latent inversion, not an incident.

## What replaces it

| New | Covers |
|---|---|
| `efteling/__tests__/eftelingStatus.test.ts` | real `mapState`, nine WIS states plus the unknown-state warning branch and a missing state, neither reachable from the table |
| `sixflags/__tests__/sixflagsStatus.test.ts` | real `mapStatus`, including the second argument the table could not express: for an empty status a zero wait is the absence of a reading, not a walk-on |
| `hfe/__tests__/hfeStatus.test.ts` | real mapper, with the `parkIsOpen` gate the old table got backwards |
| `parcsreunidos/__tests__/timeRange.test.ts` | all four label formats, plus the property the inversion broke: no format may return a close at or before the open |

HFE had no test directory at all and its mapping lived inside `buildLiveData`, so it is extracted into an exported `mapHfeRideStatus`. **Pure code motion**, checked against the original chain over 2106 input combinations with zero differences.

The timezone block was the genuine one. Its two DST-boundary cases are not covered by `datetime.test.ts` and move there; the rest duplicated what is already asserted against `constructDateTime`. The HTML block duplicated `htmlUtils.test.ts`, which already covers the same inputs against the real `decodeHtmlEntities`.

## Verification

- Full suite: 119 files, 2628 tests, passing. Net +14 after removing 43 self-asserting tests and adding 57 real ones.
- `npm run typecheck` clean.
- HFE equivalence: 2106 combinations, 0 mismatches.
- `parseTimeRange` re-probed on all six label shapes; the four that were already correct are unchanged.


cubehouse:
## Reviewed on four axes; one finding reversed a change in this PR

### The parser fix is withdrawn

`parcsreunidos.ts` is back to `main`. I justified the Dutch afternoon roll with "latent, not live" after sampling Bobbejaanland and finding 0 inverted rows in 180. **Bobbejaanland is the one park in the module that structurally cannot invert** — every time-bearing label it publishes is already 24-hour. I picked the single park guaranteed to return zero.

The module publishes **98 inverted rows right now, 93 of them `OPERATING`**, all from the 24-hour branch this PR never touched:

```
Movie Park Germany           199 entries    0 inverted
Bobbejaanland                180 entries    0 inverted
Mirabilandia                 148 entries    2 inverted
Parque de Atracciones Madrid 382 entries   13 inverted
Parque Warner Madrid         245 entries   83 inverted
```

The roll was independently wrong. `19 tot 1u` became `19:00 -> 13:00`, still inverted and with the real 01:00 destroyed; `10 tot 0u` became a plausible `10:00 -> 12:00` where it had been a visibly broken `10:00 -> 00:00`; an afternoon `2 tot 8u` stayed broken because the roll only touches the close. And the rule came from a label that appears in no live feed — its only attestation was the fixture in the deleted test file, i.e. a test that never ran shipped code. Inferring production behaviour from an invented example, in the change set written to stop exactly that.

Carried to its own card with the correct fix: one next-day rule at the caller covering all four branches.

`timeRange.test.ts` is now a characterisation test pinning today's behaviour, inversions included and labelled KNOWN BROKEN, so the real fix has to update them.

### The tautology is gone

Three reviewers independently flagged `expect(X).toBe(X)` on an interpolated string. It went with the rewrite.

### A latent CI failure I introduced

The `addDays` DST test **failed under `TZ=America/New_York`** and passed here only because this host is Europe/London. It also documented the opposite of the contract: `addDays` is `setDate`/`getDate`, local calendar arithmetic, so a mutation rewriting it to absolute arithmetic survived the entire suite. Replaced with a zone-pinned test asserting the real contract, verified to kill that mutation in both zones. The `addMinutes` case was dropped as a duplicate with misleading framing.

Full suite now green under Europe/London, America/New_York, Asia/Tokyo and Australia/Sydney.

### Smaller fixes

- The extracted mapper had been inserted between Dollywood's doc comment and the Dollywood class, so the class lost its documentation. Moved into its own section above the Park Subclasses banner; the Dollywood region is byte-identical to `main` again.
- The HFE case-insensitivity test passed against a case-sensitive implementation, because a lowercase closure reaches CLOSED through the fallthrough for the wrong reason. Added a lowercase delay, which discriminates.
- Extended closure-beats-a-posted-wait to all three closure statuses.
- Pinned three reachable, previously untested sharp edges: the status is never trimmed, `UNDER` is a substring match, and the delay branch outranks the display branch.
- Noted that the Six Flags OPERATING rows are vacuous with respect to the branch they name, since the fall-through default is also OPERATING.

### What held up

The HFE extraction survived. A reviewer rebuilt the original chain and compared whole `LiveData` objects across **150,150 combinations** including non-string inputs, `NaN`, `±Infinity` and thrown errors: zero differences. Independent mutation testing caught 27 of 31 injected faults across the four park suites, including the specific bug each was written for.

### One reviewer claim I did not take

It reported that Warner's Spanish `"12:00 a 20:00"` range is unparsed and silently dropping a session on 90 days. The code asymmetry is real — `extractLabelDescription` has an explicit `ES` branch and `parseTimeRange` has none — but I found no label of that shape in either Madrid park's live calendar today. Recorded as an asymmetry rather than an observed failure.

